PR #11, opened 2017-09-16
File drop overlay

jakedex:
- Button component: div -> button
- File drop overlay
- Minor visual fixes

### DIFF
--- a/components/ReadFileDropContainer.js
+++ b/components/ReadFileDropContainer.js
@@ -14,51 +14,30 @@ const spec = {
 
 const collect = (connect, monitor) => ({ connectDropTarget: connect.dropTarget(), isOver: monitor.isOver() })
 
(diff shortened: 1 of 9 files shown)


components/ReadFileDropContainer.js
```diff
@@ -12,8 +12,53 @@ const spec = {
...
+const collect = (connect, monitor) => ({ connectDropTarget: connect.dropTarget(), isOver: monitor.isOver() })
 
-const ReadFileDropContainer = props => props.connectDropTarget(props.children)
+class ReadFileDropContainer extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { isFileOver: false }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
mfix22: You should not need to set this state right? Once the props are updated it should rerender...

PR #14, opened 2017-09-20
Twitter hash in carbon copy tweets

mfix22:
Easter 🥚s


briandennis:
Probably doesn't matter, but do you think we should just use the `mediaId` or `tweetId`?

mfix22:
@briandennis should i slice it to 8 characters or just keep it as is?

briandennis:
✂️

PR #21, opened 2017-09-25
Add F# / OCaml mode

Krzysztof-Cieslak:
Adds F# and OCaml support by using CodeMirror's `mllike` mode - https://github.com/codemirror/CodeMirror/tree/master/mode/mllike


mfix22:
@Krzysztof-Cieslak thanks so much! Post some OCaml code here once you see it live! Merging.

PR #23, opened 2017-09-25
Use production build of React

elrumordelaluz:
Switch the temp strategy in `next.config.js` to use the production version of React from [here](https://github.com/zeit/next.js/issues/1195#issuecomment-326901737)

To avoid this 👇
screen shot 2017-09-25 at 18 13 45

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "react-spinner": "^0.2.7",
     "react-syntax-highlight": "^0.0.6",
     "tohash": "^1.0.2",
-    "twitter": "^1.7.1"
+    "twitter": "^1.7.1",
+    "uglifyjs-webpack-plugin": "1.0.0-beta.1"
   },
   "now": {
     "env": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,5 +44,6 @@
       "TWITTER_ACCESS_TOKEN_KEY": "@twitter-access-token-key",
       "TWITTER_ACCESS_TOKEN_SECRET": "@twitter-access-token-secret"
     }
-  }
+  },
+  "devDependencies": {}
```
mfix22: Not trying to be too picky, but can you remove the `devDependencies` line?

mfix22:
Awesome thanks so much @elrumordelaluz! Merging.

package.json
elrumordelaluz: oh right, sorry for this empty line, in fact I'd have done `yarn --dev add uglifyjs-webpack-plugin@1.0.0-beta.1` instead of `yarn add uglifyjs-webpack-plugin@1.0.0-beta.1`

mfix22: No problem. We will take care of that before the next release.

PR #27, opened 2017-09-25
Add Nim language

dom96:
Please verify that this is supported, you use highlight.js right? AFAIK it supports Nim :)

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -187,6 +187,10 @@ export const LANGUAGES = [
     module: 'nginx'
   },
   {
+    name: 'Nim',
+    module: 'nim'
+  },
+  {
     name: 'Pascal',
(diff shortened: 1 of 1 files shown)


mfix22:
Actually we rely on the Codemirror modes. If you check [here](https://unpkg.com/codemirror@5.26.0/mode/nim/nim.js) you can see that Nim is not supported. 

Something you could do is:
```javascript
{
    name: 'Nim',
    module: 'javascript'
}
```

or some other language that is similar to Nim (I am not familiar with it personally 😄 )

dom96:
Aww, that's a pity. I don't really want to enable poor highlighting for Nim. :)

Thanks anyway.

mfix22:
@dom96 good news with PR #35 now you can!

PR #35, opened 2017-09-26
Add Nim language

briandennis:
Hey @dom96, I was able to find a custom Nim codemirror plugin that I think is working. I'm not too familiar with Nim though. Would you mind testing some examples with the language set to `Nim` at [this url](https://carbon-jesazyjrfd.now.sh/) and seeing if they highlight correctly?

references #27

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -188,7 +188,8 @@ export const LANGUAGES = [
   },
   {
     name: 'Nim',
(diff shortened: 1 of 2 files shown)


dom96:
Looks good :D

briandennis:
Awesome, thanks so much for your help!

PR #58, opened 2017-09-28
Use CodeMirror Kotlin

andrewda:
Looks like CodeMirror already had a Kotlin mode ready to go, it was just hidden under some MIME types. This gets rid of the unnecessary `kotlin.js` custom mode, and allows for the addition of more languages that need MIME types (Java, C#, C++, Scala, etc.).


mfix22:
What if we just configure MIME modes like:
```javascript
{
  "name": "Kotlin",
  "mime": "text/x-kotlin"
}
```
and them import as `require(language.module || language.mime)`?

mfix22:
Actually, now I see what is happening. MIME types don't need to be imported. @briandennis how do we want to distinguish types like this?

briandennis:
I think the easiest way to do this might be to have MIME and module since you can have the case where a MIME's corresponding mode may not already be imported by another language.

We should also probably rename `module` field to `mode` to avoid confusion.

```javascript
{
  name: "Kotlin",
  mode: "clike",
  mime: "text/x-kotlin"
}
```
**Logic for `require`ing modes:** keep a set of modes that have already been required, don't require it again if it's already been added

**Logic for setting language in CodeMirror component:** If language has a mime, pass that. Otherwise, pass the mode.

andrewda:
How will we deal with the auto detected language being `kotlin` when the mode is `clike`?

andrewda:
Alright here's a summary of the changes I've made:

- Use `mode` instead of `module`
- If `mime` is defined, use that instead of `mode`
- Only require modes once
- For auto detection of languages, add a `short` property if the `mime` makes the language different from the `mode`. The `short` property will be used if no language could be found searching the `mode`s.

PR #71, opened 2017-09-30
Update yarn lock file

fernahh:
**Problem**

On https://github.com/dawnlabs/carbon/commit/188334db6daa99d1eaeb8f2a5126c550c4e3e175, `cross-env` was added as a dev dependency, but it's not present on `yarn.lock`. When I run `yarn start` (as indicated on readme), the lock file was updated.

**Solution**

Just a commit with the update.


jakedex:
Good catch! Thanks!

PR #77, opened 2017-09-30
Fix contribute section

fernahh:
**Problem**
On the [contribute section](https://github.com/dawnlabs/carbon#contribute), there is a instruction to start application, but this command is wrong. The correct is [`yarn dev`](https://github.com/dawnlabs/carbon/blob/master/package.json#L7).

**Solution**
Change `yarn start` to `yarn dev`.


mfix22:
Hey @fernahh thanks so much! Actually we need to remove the `yarn build` then. `yarn dev` handle both the build and start (:

mfix22:
Thank you @fernahh!

PR #78, opened 2017-09-30
Add more JavaScript-like languages

andrewda:
This adds `JSON` and `TypeScript` to this list of languages.

### JSON
json

### TypeScript
typescript

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -182,6 +182,12 @@ export const LANGUAGES = [
     mode: 'javascript'
   },
   {
+    name: 'JSON',
+    mode: 'javascript',
+    mime: 'application/json',
+    short: 'json'
+  },
+  {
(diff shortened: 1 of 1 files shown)


mfix22:
Super cool. @andrewda lets maybe open an issue for languages to remove? Just so the list doesn't grow to unwieldy lengths (: Thanks for this one though, definitely needed!

andrewda:
@mfix22 One way to go about that problem is adding a text box to the top of the language list and being able to search for a specific language. That way we can include as many languages as possible without having to search through a long list.

mfix22:
@andrewda sounds like https://github.com/paypal/downshift could be a perfect solution. There are tons of examples that have match-sorting for drop downs. 

@jakedex what do you think about this, design wise

PR #82, opened 2017-09-30
React-codemirror -> react-codemirror2

### LINKED ISSUE #64: Use `react-codemirror2` instead of `react-codemirror`
[`react-codemirror2`](https://github.com/scniro/react-codemirror2) is much more up to date than [`react-codemirror`](https://github.com/JedWatson/react-codemirror), which is still on a version of CodeMirror that's more than 1 year old.

mfix22:
- Closes #64


mfix22:
Thanks @andrewda

Issue #79, opened 2017-09-30
Create build script to add contributors to the README


mfix22:
Similar to all-contributors, or open-collective, but personalized for Dawn

We want a way to recognize all our awesome contributors!

yeskunall:
I got you guys with this one 💪 expect a PR soon enough!  ❤️

PR #87, opened 2017-10-01
Getting started docs

jkling38:
Adding instruction to run npm install as it may not be obvious to people new to NodeJS development.


CONTRIBUTING.md
```diff
@@ -4,5 +4,6 @@ If you have discovered a bug or have a feature suggestion, feel free to create a
 
 If you'd like to make some changes yourself, see the following:
 1. [Fork](https://help.github.com/articles/fork-a-repo/) this repository to your own GitHub account and then [clone](https://help.github.com/articles/cloning-a-repository/) it to your local device
-2. Build and start the application: `yarn dev`
-3. Finally, submit a [pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) with your changes!
+2. Run `npm install` to download required packages.
```
mfix22: I agree! Can you actually change this to include a step about needing `yarn` (npm i -g yarn) and then change `npm install` to `yarn`.

mfix22:
@jkling38 thanks for this! Definitely agree with this. Also if you want to help with #79 that would be awesome!

jkling38:
@mfix22 I can give the contributors build script a shot. I'm brand new to next js so it may take a little while. I assume you want yarn build to check README.md for the contributor's username and if it isn't already there then add it as a link with the full name?

mfix22:
Actually no, I just meant an entirely separate script, not related to Next or `yarn build`, that we can either run before a `commit` or a `push` to add contributors to the README. Like `yarn contributor` runs `node ./scripts/contributors.js`, or something similar. If this isn't something you are passionate about please skip with what you like (:

Issue #75, opened 2017-09-30
Read content & preferences from URL query string


jgierer12:
This would make sharing links to the page much easier.

For example, `https://carbon.now.sh/?code=Hello%20world!&theme=blackboard` would pre-fill the editor with "Hello world!" and apply the *Blackboard* theme. Everything else would fall back to the default settings, of course.

If this would be a welcome addition, I'll open a PR for it.

andrewda:
I think the issue with this is mainly the length of the code – we've only got 2048 characters to work with, minus the theme. One thing we could do is use [base65536](https://github.com/qntm/base65536) to shorten that code significantly, but it leads to ugly-looking URLs.

andrewda:
For example,

```javascript
const hello = () => {
  console.log('Hello World!');
}
```
(56 chars)

Encodes to: 
`堗臼铤⋆凃堜曚䙨㫐㖏橯圂Ⴃ㇝荆闌壷䅛鍅Ꮄ榋堑蒠総徖㽈䬥㧰穏` (29 chars)

Just for fun, I also tried encoding the `server.js` file (1077 chars). It encodes to:

```
[...]

Issue #52, opened 2017-09-27
Carbon CLI


jakedex:
If there's enough support, we could create a CLI interface for Carbon.

andrewda:
I'm super interested in working on a CLI. How would it work, though? The only way I can see it working at the moment is using an API from Carbon to fetch the image. The downside to this is that the user would need an internet connection. An alternative would be for the user to self-host Carbon, but that would result in high startup times.

It'd be great if we could think of a way to do this without a server.

jakedex:
https://github.com/dawnlabs/carbon/pull/88#issuecomment-333429240

briandennis:
@andrewda the functionality I'm working on now to get the app working in Safari should be pretty similar to what's needed for the CLI locally. Once I finish that, let's chat

itaditya:
Hey guys I have made a simple CLI here https://github.com/itaditya/carbon-cli . Take a look https://www.npmjs.com/package/carbon-cli

monouser7dig:
> If there's enough support, we could create a CLI interface for Carbon.

as a bash script or alike without npm? would be very welcome and sleek 

like `carbon --syntax=c --theme=material input.c output_c.png`

zmwangx:
Maybe it's possible to build an entirely local CLI with puppeteer?

mfix22:
Hey all, issues marked "Integrations" are being closed and theirs discussions moved to https://github.com/dawnlabs/carbon/issues/358. With many requests for new features, this allows the Carbon team to prioritize which to support first 👍

Issue #48, opened 2017-09-27
Editor Integrations


jakedex:
We should create simple plugins for Atom, VS code, Sublime with `select text -> command palette -> create screenshot` support. 

(source: https://www.reddit.com/r/webdev/comments/72qz1k/carbon_create_and_share_beautiful_images_of_your/dnl14fp/)

jakedex:
Closing as these have been created — https://github.com/dawnlabs/carbon#community

PR #88, opened 2017-10-01
[PoC] Preconfigured Editor

zcei:
In order to enable features like #48 and #52 (and potentially #75, once it's clear how to combine localStorage and query string settings), there is a need to use this app in a more "programmatic" approach than currently possible.

For now I added a new endpoint, which is a "bare minimum version" of the editor. Ideally the shared functionality could be abstracted and reused by both editors.
It could also be extended to display the settings and update the query string whenever a setting changes.

Most important difference to the current editor is `componentWillMount`, which exposes two methods on the `window` object, which can be used to instrument the editor.

I built a [small server](https://github.com/zcei/carbon-server) which makes use of those methods to render the images on server side: 

```sh
curl \
  -X POST \
  -d 'const foo = bar' \
  https://carbon-server-ezesmrobex.now.sh/?wc=false&theme=ambiance
```

⚠️ This PR is not intended to be merged immediately, but to share my state early in order to gather feedback.

### DIFF
--- a/server.js
+++ b/server.js
@@ -21,6 +21,7 @@ app.prepare().then(() => {
   server.use(morgan('tiny'))
 
   server.get('/about', (req, res) => app.render(req, res, '/about'))
+  server.get('/preconfigured-editor', (req, res) => app.render(req, res, '/preconfigured-editor'))
 
   // if root, render webpage from next
   server.get('/*', (req, res) => app.render(req, res, '/', req.query))
--- a/lib/util.js
+++ b/lib/util.js
@@ -11,3 +11,54 @@ const parse = v => {
 
 export const getState = morph.compose(parse, morph.get(KEY))
 export const saveState = (window, v) => assign(window, JSON.stringify(v))
+
+export const defaults = (value, defaultValue) => {
+  if (typeof value === 'undefined') {
+    return defaultValue
+  }
+
+  return value
+}
+
+const asBoolean = value => (value ? value === 'true' : true)
+const identity = value => value
+
+export const queryToProps = function(query) {
+  const mappings = {
+    bg: {
+      key: 'background',
+      transform: value => (value.startsWith('#') ? value : `#${value}`)
+    },
+    theme: 'theme',
+    lang: 'language',
+    shadow: {
+      key: 'dropShadow',
+      transform: asBoolean
+    },
+    wc: {
+      key: 'windowControls',
+      transform: asBoolean
+    },
+    vp: 'paddingVertical',
+    hp: 'paddingHorizontal'
+  }
+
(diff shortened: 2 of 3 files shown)


server.js
```diff
@@ -21,6 +21,7 @@ app.prepare().then(() => {
   server.use(morgan('tiny'))
 
   server.get('/about', (req, res) => app.render(req, res, '/about'))
+  server.get('/preconfigured-editor', (req, res) => app.render(req, res, '/preconfigured-editor'))
```
andrewda: This is pretty long for a path – maybe shorten it to something like `/configured`?

mfix22:
@zcei I definitely agree with this concept. Basically turning the image creating code into an API. We already have an express server running Carbon, so could you integrate carbon-server into the server we already have in this code base (we cannot depend on a Carbon server that we don't have full deploy control over). 

This is an awesome start.

zcei:
> turning the image creating code into an API

Sadly it's not as easy as this, because for proper rendering all kind of computations have to be made, which currently involves spawning a headless Chromium via [`puppeteer`](https://github.com/GoogleChrome/puppeteer).

> could you integrate carbon-server into the server we already have

Honestly, I don't think this makes sense. The overhead of building the Docker image to include Chromium and all this stuff would massively slow down frontend related work.
Also it feels weird to me to call a server endpoint, which spawns a browser page that in turn visits the same server. A clear separation of concerns makes it easier to grasp imho.

> we cannot depend on a Carbon server that we don't have full deploy control over

Well, I wouldn't mind handing over the repo to you, if that's the sole reason.

jgierer12:
> Honestly, I don't think this makes sense. The overhead of building the Docker image to include Chromium and all this stuff would massively slow down frontend related work.

I'm not an expert on this, but wouldn't something like [chromeless](https://github.com/graphcool/chromeless) work (with Chrome running on AWS Lambda)

zcei:
Yes, this would totally work. 
Actually there are [plans to collaborate](https://github.com/graphcool/chromeless/issues/235#issuecomment-323359507)

briandennis:
Is there any reason this couldn't be accomplished locally rather than over a networked API? I don't know much about editor plugins, but for a CLI implementation I think we could definitely build a node program to use headless chrome and render the image on the local machine rather than ping a server somewhere.

yeskunall:
I agreed with @briandennis. The CLI (and the CLI only) should avoid making a network request, if possible. We should first try by running a local instance of the app, combined with [`puppeteer`](https://github.com/GoogleChrome/puppeteer) and see how that goes.

mfix22:
Gonna close this since a couple months have gone by without discussion.

PR #89, opened 2017-10-01
Lazy load CSS files

### LINKED ISSUE #81: Lazy load stylesheets for themes

mfix22:
Closes #81

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -6,10 +6,7 @@ import Spinner from 'react-spinner'
 import toHash from 'tohash'
 import WindowControls from '../components/WindowControls'
 import CodeMirror from '../lib/react-codemirror'
-import { COLORS, DEFAULT_LANGUAGE, LANGUAGES } from '../lib/constants'
-
-const LANGUAGE_MODE_HASH = toHash(LANGUAGES, 'mode')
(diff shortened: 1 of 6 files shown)


mfix22:
@jakedex there is a bug currently where if you load state from local storage, the state gets updated appropriately and the theme will be displayed in the CodeMirror instance, but the dropdown selected title does not change (still shows "Seti", for example). Can you take a look and push a fix to this branch?

briandennis:
So this works by loading the CSS at the time when the theme is switched?

mfix22:
@briandennis yeah. Should we handle this? Will up time-to-first-paint.

briandennis:
Yeah, I think it may make more sense to first load either the default theme or the query param theme, then paint, then load all remaining themes. That may be 1) simpler and 2) wouldn't lead to lag on theme switching while the style sheet loads. 

The advantage of your way though is that it only loads the ones you need, so for use cases where total data used matters (like mobile) it would be beneficial. Maybe in the future we can sniff the user agent and use this technique if they're on a phone?

mfix22:
Closing this. Unnecessary for now: 
- Closes #81 

screen shot 2017-10-04 at 8 08 59 pm

PR #90, opened 2017-10-02
Added button hover styling

SAXTEN2011:


mfix22:
@SAXTEN2011 
1. Can you add `.idea` to the .gitignore and remove those files from this PR?
2. Can you post a screenshot showing the fix? Thanks!

yeskunall:
IMO, the `.gitignore` should only contain project-specific files we'd want to ignore. Stuff like IDE/editor configuration files should not be pushed to remote. They should instead be a part of every developer's `.gitignore_global`.

If @SAXTEN2011 is really interested in getting this PR merged, and if what I said makes sense, then [here's](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) how you would set up a global `.gitignore` locally.

Otherwise, carry on, fellas! 🚶

jakedex:
Thanks for the PR, @SAXTEN2011! We decided to go a different direction with the hover styles in #96. Closing.

Issue #83, opened 2017-10-01
button apperance change on hover


PAPERPANKS:
![changing color of button on hover](https://user-images.githubusercontent.com/13312568/31052870-5e812ca2-a6ad-11e7-97c1-64073de462b0.png)

```
button:hover {
    background: props.color;
    color: #fff;
    border: 0.5px solid #fff;
}
```

PR #93, opened 2017-10-02
added button hover and adjusted colorpicker alignment

manask4:
Hi, I have just added button hover effects and minor alignment on the colorpicker block.
I've attached a few images here.

Colorpicker block before
bgcolorbefore
Colorpicker block after
bgcolorafter

No hover effect
nohovercolor
With hover effect
withhovercolor

I think this addresses the issue [#83](https://github.com/dawnlabs/carbon/issues/83).

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -1,33 +1,60 @@
 import React from 'react'
 import { COLORS } from '../lib/constants'
 
-export default props => (
-  <button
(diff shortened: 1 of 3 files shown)


jakedex:
Thanks for the PR, @manask4! We decided to go a different direction with the hover styles in #96. Closing.

PR #94, opened 2017-10-02
Fix failing CI

andrewda:
Looks like https://github.com/dawnlabs/carbon/commit/0c365218d67508c992c080cde1ec293fbca0a953#diff-11a82058b621f771e497118318cf203fR75 introduced a tiny issue that Prettier didn't correct, so the CI is failing. A few open PR's already have this change, but in case those take a while to get merged in, this explicitly fixes it.

### DIFF
--- a/pages/editor.js
+++ b/pages/editor.js
@@ -72,7 +72,7 @@ class Editor extends React.Component {
     saveState(localStorage, s)
   }
 
-  getCarbonImage () {
+  getCarbonImage() {
     const node = document.getElementById('section')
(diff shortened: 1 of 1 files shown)


yeskunall:
This was fixed in #93 (I _think_, correct me if I'm wrong tho), but yeah in case that PR takes a while to merge, your change should effectively stop the build from failing.

andrewda:
@yeskunall Yep I believe it was. If that get merged in first, great!

PR #95, opened 2017-10-03
Use spread operator

neighborhood999:
- Replace `Object.assign()` to spread operator
- Fix meta tag: `http-equiv` to `httpEquiv`


mfix22:
Cool with me! Thank you!

Issue #92, opened 2017-10-02
Use `.github` for GitHub templates


yeskunall:
Maybe create a `.github/` folder at this point, and move `CONTRIBUTING.md` inside it? Also, maybe create an `ISSUE_TEMPLATE.md` and `PULL_REQUEST.md`?

But most importantly, in lieu of #79, I was thinking that maybe we can add a `CONTRIBUTORS.md` inside `.github/` and link it in `README.md` instead? This way the README itself doesn't have to get cluttered unnecessarily.

jakedex:
Thanks for tackling this one, @yeskunall! I'm definitely on board with adding a `.github` folder, but I think we want to keep our contributors section in the `README.md` in order to keep it visible as possible. I'm all for moving it into a `CONTRIBUTORS.md` and linking down the road if it gets too cluttered. 

See https://github.com/kentcdodds/all-contributors#mandatory

yeskunall:
Sure thing. This will make the build script a bit more tricky than it is right now, but I'll show you what I guys did in the PR.

Once that's approved, a PR for creating a `.github` folder will be up as well.

Cheers! ❤️

yeskunall:
At work. No worries tho, this will definitely have a PR by today evening. Should we also introduce basic templates for issues and PRs?

jakedex:
No rush! And yeah, let's add templates for both.

PR #100, opened 2017-10-03
Add a build script to add contributors to the README

### LINKED ISSUE #79: Create build script to add contributors to the README

yeskunall:
So I quickly put this together (hence the **WIP** in the title), because I want feedback before I make any drastic additions. Currently, it just adds a raw output (formatted to Markdown) at the end of the README, which works fine if it's the first time. But the next time it's run, it just appends the result to the end of the file, which isn't what we're looking for.

So my question is this: since @jakedex wants to add the contributors to the end of the `README` for now (as discussed in #92, should we just manually add them from time to time? When time comes, we can use something like [all-contributors-cli](https://github.com/jfmengels/all-contributors-cli) to generate a `CONTRIBUTORS.md` for us and it will (hopefully) handle the issues I've pointed out here.

... or should I add some logic to the current script to check for existing users, and only add those that weren't previously in the list?

Thoughts?

- Closes #79


jakedex:
Thanks for getting the ball rolling on this. It looks like [all-contributors-cli](https://github.com/jfmengels/all-contributors-cli) will handle injecting updates to the `README.md`, let's just add a `.all-contributorsrc`, a npm script for `contrib:add`, and `contrib:build` to `precommit`?

yeskunall:
Whoopsie, should have read their README a bit more closely. 🙊 if that's the case, I'll look into it right away!

yeskunall:
Hey @jakedex 👋

I just wanna make sure we're on the same page: you want a `contrib:add` that basically calls `all-contributors add`, which would mean a contributor is added using the CLI? How would that work?

... and I'm guessing `contrib:build` was to be followed by the previous command that ran `all-contributors generate`?

In any case, look at the changes I've introduced. It works just fine, except that there's no way to add emoji keys to the contributors as I'm just pulling in the contributors, not the _type_ of contribution (which I don't think is possible)

mfix22:
@yeskunall thanks for this! Now that we are going with all-contributors, you can just use `all-contributors add` and `all-contributors generate`.  [See these lines for reference](https://github.com/kentcdodds/all-contributors/blob/master/package.json#L16-L17).

And actually, the emojis are generated from the key words in the contributions array. You can see an example here: https://github.com/kentcdodds/all-contributors/blob/master/.all-contributorsrc#L15-L19

jakedex:
Apologies for the vague response, @yeskunall. I was trying to say that we should just do this all manually with the npm scripts and scrap `contributors.js` for the time being. Again, sorry for the confusion!

yeskunall:
All right, understood. I'll scrap the script and manually add the contributors, update the README. Gotta leave for work, but this should be merged by today evening.

❤️

yeskunall:
All right, so the npm scripts are done and good to go. Final question: @mfix22, who's going to add the emojis keys to the contributors' block? 😛

I'm willing to do it, but the CLI asks questions of the likes:

capture

... and I'm not sure if I should be the one deciding that. 🚶

(Once I get a response, I'll also fix the npm `precommit` script. Noticed it after I pushed 🙈

mfix22:
I can handle that @yeskunall (: mind if i run the CLI and push to your branch?

yeskunall:
Be my guest! ❤️

> --

Kunall Banerjee

mfix22:
Thanks @yeskunall for all your work and comments!

PR #111, opened 2017-10-08
Save and update query params based on editor state

### LINKED ISSUE #75: Read content & preferences from URL query string

mfix22:
Update editor to save and load editor state from query parameters (e.g. `?code=Hello&l=javascript`)

Please help review and test 😄 

- Closes #75


pages/editor.js
```diff
@@ -23,51 +24,64 @@ import {
...
+        dropShadow: true,
+        windowControls: true,
+        widthAdjustment: true,
+        paddingVertical: '48px',
+        paddingHorizontal: '32px',
+        uploading: false,
+        code: props.content,
+        _initialState: this.props.intialState
+      },
+      this.props.intialState
```
yeskunall: Could just use the spread syntax here.

pages/editor.js
```diff
@@ -23,51 +24,64 @@ import {
...
+        theme: THEMES.seti.id,
+        language: DEFAULT_LANGUAGE,
+        dropShadow: true,
+        windowControls: true,
+        widthAdjustment: true,
+        paddingVertical: '48px',
+        paddingHorizontal: '32px',
+        uploading: false,
+        code: props.content,
+        _initialState: this.props.intialState
```
yeskunall: Typo: should probably be `initialState`. L62 and L64.

mfix22: Great catch! I might have never seen that!

pages/editor.js
```diff
@@ -23,51 +24,70 @@ import {
...
+    if (!this.state._initialState) {
+      const state = getState(localStorage)
+      if (state) {
+        this.setState(state)
+      }
     }
   }
 
   componentDidUpdate() {
+    const mappedState = mapper.map(reverseMappings, this.state)
```
briandennis: This should probably be done inside of `replaceQuery` so all that the editor component needs to do is pass a state object to a function. Also maybe rename to `updateQueryString` or something to make its use clear?

briandennis: Or another way, editor probably shouldn't have to know about `mapper` or `reverseMappings`

PR #115, opened 2017-10-19
Fix README

yeskunall:
## Description

Carbon actually expects the URL to be like `/YOUR_USERNAME/GIST_ID_HERE`

Example URL: https://carbon.now.sh/yeskunall/0d2a1e461bb67114133c

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Visit [carbon.now.sh](https://carbon.now.sh) or read [our post](https://dawnlabs
 
 There are a few different ways to import code into Carbon:
 - Drop a file onto the editor
-- Append a GitHub gist id to the url  (e.g. `carbon.now.sh/GIST_ID_HERE`)
+- Append a GitHub gist id to the url  (e.g. `carbon.now.sh/YOUR_USERNAME/GIST_ID_HERE`)
 - Paste your code directly
 
 #### Customization


andrewda:
Both work AFAIK:

For public gists, these are fine:
https://carbon.now.sh/f8bf5e0e093c142e0b12877dfd64d468
https://carbon.now.sh/andrewda/f8bf5e0e093c142e0b12877dfd64d468

And for secret gists, these both work:
https://carbon.now.sh/b18e33b9368e0579ee9cc7599ab2d19f
https://carbon.now.sh/andrewda/b18e33b9368e0579ee9cc7599ab2d19f

yeskunall:
Can you try it with [this](https://gist.github.com/yeskunall/0d2a1e461bb67114133c) gist @andrewda? I might be doing something wrong :thinking:

andrewda:
It _might_ be because GitHub started using a new format for their gist URL's. Looks like the newer ones (like the ones in my comment) are uniquely identifiable in the GitHub database (https://gist.github.com/b18e33b9368e0579ee9cc7599ab2d19f `301`'s to https://gist.github.com/andrewda/b18e33b9368e0579ee9cc7599ab2d19f). For the older links, I'm guessing they're per-user.

As for what the documentation should say, maybe just something like:

> - Append a GitHub gist id to the url  (e.g. `carbon.now.sh/GIST_ID_HERE` or `carbon.now.sh/USERNAME/GIST_ID_HERE`)

to make it clear that both work.

mfix22:
Both work. We actually just split on `/` and grab the last value, using it as the Gist ID in the fetch.

In fact, you can just do this if you want:
`https://carbon.now.sh/https://gist.github.com/mfix22/acba311bfced1924d78ccbb6a0a5ad3c`

although it is a little much 😄 

Thanks for the PR but we are gonna stick with how we have it.

andrewda:
@mfix22 Hmm, interesting. Any ideas as to why https://carbon.now.sh/yeskunall/0d2a1e461bb67114133c works but https://carbon.now.sh/0d2a1e461bb67114133c does not?

yeskunall:
Same question as @andrewda. 🙋 in fact, I made the PR specifically because I noticed it didn't work in my case.

mfix22:
@yeskunall and @andrewda thanks for the bug catch! It was just a matter of the fact that old Gist ID lengths were shorter than 32 characters, and I had not accounted for that. This commit should fix the issue once it is deployed: 1277a1114384264e198c95f856fef43d05cec61f.

The way we check to see if a Gist should be fetched or not is still hacky, and could be improved (PR anyone?) 😄 

Thanks again! Closing.

PR #132, opened 2017-10-31
fix cursor jumping at the end after every editor update

slorber:
Currently the cursor constantly jumps at the end everytime you press a key in the editor

My changes is supposed to prevent that


slorber:
using onBeforeChange may look "surprising" but it's what's in the demo:

image

briandennis:
@slorber awesome! Thanks so much for fixing this up, really appreciate it 😀

slorber:
thanks to you @briandennis , your project was very useful inspiration for something I just published on git: https://github.com/slorber/react-reboot

PR #133, opened 2017-10-31
Fix MySQL syntax highlighting.

varbrad:
The MySQL syntax highlighting does not currently work unless you explicitly define the correct MIME type associated with it (`text/x-mysql`).

I have added an extra syntax option, SQL (MySQL), which adds the MIME type to the object and this now makes MySQL-flavoured SQL syntax highlighting work as expected.

__Before__
Old SQL Syntax Example

__After__
MySQL Fix Example


lib/constants.js
```diff
@@ -295,6 +295,12 @@ export const LANGUAGES = [
     mode: 'sql'
   },
   {
+    name: 'SQL (MySQL)',
```
jakedex: Could you rename this to MySQL?

varbrad:
Renamed are re-positioned to keep alphabetical ordering of the languages in the drop-down. 👍

jakedex:
Thanks again for this, @varbrad! We'll have it deployed soon here.

PR #141, opened 2017-11-13
Debounce language changes

### LINKED ISSUE #140: Wait a while before checking language in auto
When deleting or typing out lots of code at once, Carbon will try to guess the language after each key stroke, which takes some time and adds latency before updating the rendered text. This can lead to an unresponsive feel if trying to type quickly, especially on slower computers. We might consider only checking the language in auto mode after 200ms or so idle. [This component](https://github.com/reacttraining/react-idle) might be useful, though it also wouldn't be tough to write our own idle timer.

mfix22:
- Closes #140

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -4,6 +4,8 @@ import React from 'react'
 import domtoimage from 'dom-to-image'
 import Spinner from 'react-spinner'
 import toHash from 'tohash'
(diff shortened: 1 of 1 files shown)


mfix22:
@andrewda does this work for you? Should we increase the timeout threshold?

PR #148, opened 2017-12-15
Collapsible "Advanced" settings

mfix22:
screen recording


jakedex:
Hell yes, thanks for cleaning this up. Could you add a dropdown arrow to the Collapse element? Also, how do you feel about changing 'width adjustment' to something like 'auto width'?

Issue #123, opened 2017-10-23
Frosted glass terminal style


jakedex:
<img width="613" alt="bbaf728e-c31a-11e6-972f-0995e77a32a0" src="https://user-images.githubusercontent.com/10369094/31911790-ab8c5f6c-b7f6-11e7-9ee9-582752a44dde.png">

![dmgz16nwaaapyj8](https://user-images.githubusercontent.com/10369094/31911846-d279ed42-b7f6-11e7-9db1-9d8b23f5126f.jpg)

- Add terminal transparency slider to settings
- This should be rather trivial to implement with css: https://css-tricks.com/frosting-glass-css-filters/

astronomersiva:
I would love to take a dab at this if @vishesh-dembla is no longer working on this.

On first look, it looks like the background style for the code editor is from CodeMirror and as such we will not be able to directly manipulate its styles.

I was thinking we could add a background property to each theme in [constants.js](https://github.com/dawnlabs/carbon/blob/master/lib/constants.js#L15) in rgb( ) and use this to override the style set by CodeMirror.

vernondegoede:
I opened a PR for a theme which comes with vibrancy support: https://github.com/dawnlabs/carbon/pull/341. It's based on the Verminal, a Hyper theme (which is in the second image) and Framer Syntax.

Hope you like it! 🎉

kissu:
@jakedex super cool idea ! :tada:

mfix22:
Closing since seems Verminal is covering the use case 👍

PR #156, opened 2018-01-17
Support adjusting window transparency

astronomersiva:
This attempts to fix https://github.com/dawnlabs/carbon/issues/123. 

screen shot 2018-01-17 at 11 51 22 pm

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -35,6 +35,8 @@ class Carbon extends PureComponent {
     this.codeUpdated = this.codeUpdated.bind(this)
     this.onBeforeChange = this.onBeforeChange.bind(this)
     this.handleLanguageChange = this.handleLanguageChange.bind(this)
+
+    this.codeMirrorNode = React.createRef();
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


components/Carbon.js
```diff
@@ -178,7 +184,7 @@ class Carbon extends React.Component {
             }
 
             #container :global(.CodeMirror-gutters) {
-              background-color: unset;
+              background-color: unset !important;
```
astronomersiva: some themes have `!important` in their background thereby rendering this useless.

lib/constants.js
```diff
@@ -13,77 +13,95 @@ export const FONTS = [
 export const THEMES = [
   {
     id: '3024-night',
-    name: '3024 Night'
+    name: '3024 Night',
+    background: 'rgba(9, 3, 0, alpha) !important'
```
mfix22: I am thinking for this you can just look at the dom-node's background color dynamically and adjust the opacity as you have done here. That way if people want to add a new theme, they can do so a lot easier than manually setting an `rgba`.

astronomersiva: Makes sense. Will make the changes (y)

astronomersiva:
Made some changes. Please have a look :)

components/Carbon.js
```diff
@@ -99,6 +99,24 @@ class Carbon extends React.Component {
       viewportMargin: Infinity,
       lineWrapping: true
     }
+
+    let background = ''
+    let { codeMirrorNode } = this
+    if (codeMirrorNode) {
+      let [codeMirrorEditor] = codeMirrorNode.ref.children
+      let backgroundColor = getComputedStyle(codeMirrorEditor).backgroundColor
```
jakedex: Are you sure this is working? It looks like the background color's always being set to black, regardless of syntax theme.

ThiHaKo:
Nice

components/Carbon.js
astronomersiva: @jakedex, sorry for the late response.

I have made some changes and pushed. The problem with the earlier change was that when a theme was changed, the background calculated from the previous theme was still being applied.

I have addressed it in today's change but I am not able to reset the transparency slider once a new theme is selected. Any ideas?

astronomersiva:
@jakedex, sorry for the late response.

I have made some changes and pushed. The problem with the earlier change was that when a theme was changed, the background calculated from the previous theme was still being applied.

I have addressed it in today's change but I am not able to reset the transparency slider once a new theme is selected. Any ideas?

PR #157, opened 2018-01-18
Implement drag-n-drop image for backgrounds

### LINKED ISSUE #122: Use image as background
I think it'd be really cool to support using user-uploaded images as backgrounds. We'd probably want to make the current color-select dropdown tabbed, with a tab for image background settings ('upload image', 'remove image', 'scale image', etc). 

- Users should be able to scale the background
- Users should be able to reposition the background (I'm thinking we could just make it draggable in the editor?)

mfix22:
- Closes #122

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -147,8 +147,13 @@ class Carbon extends React.Component {
             }
 
             #container .bg {
-              background: ${this.props.config.background || config.background};
-              position: absolute;
+              ${this.props.config.backgroundImage
+                ? `background: url(${this.props.config.backgroundImage});
+                   background-size: cover;
+                   background-repeat: no-repeat;`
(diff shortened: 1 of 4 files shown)


jakedex:
Looks good to me, one thing you should add though—when dnd'ing an image, let's change the "Drop your file here to import" copy to "Drop your image here to set it as the background".

jakedex:
Also, the image isn't persisted on reload... do we want to save the image in localStorage?

mfix22:
@jakedex I didn't want to save this keyword in local storage or the URL query because it can get pretty long. We can opt into that in a separate PR.

mfix22:
Also that's a great idea for the drop message. I'm not sure we can know which type of file it is until it's read (on Drop), but I can look into it.

PR #195, opened 2018-02-04
Fix boolean params

bdougherty:
This PR addresses an issue with boolean query parameters like window controls, line numbers, etc. They were not being parsed as booleans, so if they were set to `false` in the url, they were not being reflected properly in the rendering because `"false"` is truthy.

This PR updates the routing code to correctly parse those params to the appropriate values so they are reflected in the rendering.

### DIFF
--- a/lib/routing.js
+++ b/lib/routing.js
@@ -7,26 +7,32 @@ if (typeof window !== 'undefined') {
   history = createHistory()
 }
 
-const mapper = new Morph()
+const mapper = new Morph({
+  types: {
+    bool: v => (v ? v === true || v === 'true' : v)
+  }
+})
 
 const mappings = [
-  'bg:backgroundColor',
-  't:theme',
-  'l:language',
-  'ds:dropShadow',
-  'wc:windowControls',
-  'wa:widthAdjustment',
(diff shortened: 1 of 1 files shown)


lib/routing.js
```diff
@@ -7,26 +7,32 @@ if (typeof window !== 'undefined') {
   history = createHistory()
 }
 
-const mapper = new Morph()
+const mapper = new Morph({
+  types: {
+    bool: v => (v ? v === true || v === 'true' : v)
```
mfix22: Thanks for using the type mappings! Can you make it so that the type can only ever map to `true` or `false`? Right now a bunch `null` and `undefined` and `0` would all stay as such and we want to convert those to false.

Maybe something like:
```javascript
v => {
  if (v === 'false') return false
  return Boolean(v)
}
```

mfix22: Also please ignore Travis (:

mfix22:
Thanks so much @bdougherty

lib/routing.js
bdougherty: At first I had it that way, but it was interfering with the defaults if you go to the main url without any parameters.

PR #199, opened 2018-02-04
Added "Save as Vector (SVG)"

### LINKED ISSUE #198: Feature request: Save as Vector / PDF
I really like the way the code is rendered, but I'm a bit hesitant to include just a picture, because the code inside the image cannot be copied by others. I think it would be great if there was a possibility to download it as an SVG file or even a PDF.

martinfrancois:
Closes #198 

image


andrewda:
👏  I like it! Maybe we should restyle this later, but it might be nice to have a Save Image dropdown with a PNG/SVG option?

mfix22:
@martinfrancois thanks for hooking this up! 
1. Can you give us access to make changes to @jakedex can implement the style changes mentioned above (we like the idea of a dropdown on the button itself).

2. What do you think about passing the format  as `{ format: 'svg' }` instead? That way if we want to add more options in the future we can?

martinfrancois:
@andrewda @mfix22 you're welcome! The idea with the dropdown sounds good, feel free to implement it!
I also thought about passing it as you suggested @mfix22, feel free to change it to that as well or if you'd like I could also do it.
I think you should already have access, at least "Allow edits from maintainers" is checked.

jakedex:
Awesome! Thanks for the contribution, @martinfrancois 👍

martinfrancois:
You're welcome! Thansk for working out the details as well and restyling it 👍

Issue #1, opened 2017-06-19
Add handles for manually resizing terminal


jakedex:
https://github.com/STRML/react-resizable

jakedex:
On second thought, this doesn't seem super useful... closing for now.

PR #208, opened 2018-02-06
Support for manual resize

arjunkomath:
## Description
Implements #1 
Support for manually resizing container
Added https://github.com/STRML/react-resizable for resize support

## Known Issue
Resize handle is visible in exported image


briandennis:
Unfortunately we recently closed #1 since support for manipulating the size of the window and padding was added to the toolbar. I think that's a more intuitive way to control it and I'm not sure how the two would interact. Sorry for the confusion and thanks for taking the time to build this out!

PR #211, opened 2018-02-06
Feature/docker

SkYNewZ:
## Docker support
screenshot

## Description
- add `.dockerignore` for building without dependencies
- Dockerfile for building
- Push to registry https://hub.docker.com/r/skynewz/carbon/

To do if you are interested in this: set up an automatic deployment with each new release


briandennis:
Cool idea! For the time being we're going to leave this out — Now also supports Docker deployments so this might confuse people if we have a `Dockerfile` and still deploy as an NPM project. Maybe post  something to Docker Hub that clones the main repo? If you get that working, I'd be happy to link to it in the README like you did here.

briandennis:
That being said, we're considering possibly switching to Docker temporarily to support #213. @SkYNewZ we'd love your help there if you want to try writing a Dockerfile that supports Puppeteer!

SkYNewZ:
Okay, thank you for your answer! At the moment everything works on my side
by copying the sources from the repository, I will modify the Dockerfile to
clone rather than copy and update the current image on the Docker Hub. I'll
come back to you when everything is OK and I'll be happy to see it in
README 😁

SkYNewZ:
@briandennis Okay, I'll look into it!

PR #213, opened 2018-02-06
Safari Support

### LINKED ISSUE #28: Making save image work in all browsers
To start:
- [ ] Safari
- [x] Edge

briandennis:
## Overview
This adds support for saving images in Safari. [dom-to-image](https://github.com/tsayen/dom-to-image) uses `<foreignObject>` tags which taint the canvas and prevent conversion to a DataURL in Safari. As a workaround, this PR adds an API endpoint for generating an image based on a carbon URL. To do this, the server renders the page in chrome using [puppeteer](https://github.com/GoogleChrome/puppeteer), generates the image manually, and sends it as a response.

## Still To Do

- [x] send only Base64 encoded state not whole URL
- [x]  get puppeteer working on Now (unless [Zeit adds support for this](https://github.com/zeit/now-cli/issues/861) we'll likely have to drop down to Docker)
- [x] error handling

closes #28


SkYNewZ:
Currently I recreated the docker image that clones the repository, it works as before. However I can't test on Safari, I don't have a Macintosh on hand... :confused:
You can pull image [here](https://hub.docker.com/r/skynewz/carbon/) if you want to test.
Keep me informed!

handlers/image.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!url) res.status(400).send()
+
+  try {
+    await page.goto(url)
+    await page.addScriptTag({ path: `./lib/customDomToImage.js` })
+
+    const targetElement = await page.$('#export-container')
+
+    let dataUrl = await page.evaluate(target => {
```
mfix22: Should we pull this out into a lib function so we don't always have to update it in two places?

lib/routing.js
```diff
@@ -29,13 +29,33 @@ const reverseMappings = mappings.map(field =>
...
     .map(key => `${key}=${keys[key]}`)
     .join('&')}`
 
-export const getQueryStringState = query => mapper.map(mappings, query)
+export const getQueryStringState = query =>
+  query.state ? deserializeState(query.state) : mapper.map(mappings, query)
 export const updateQueryString = state =>
   history.replace({
-    search: encodeURI(keysToQuery(mapper.map(reverseMappings, state)))
+    search: encodeURI(keysToQuery({ state: serializeState(state) }))
```
mfix22: @briandennis doesn't this remove the human-readable query params from the output? We want that so people can understand how to use it as an API without reading the code.

briandennis: Yes, that was for testing — should be fixed. Maybe worth adding an option to get the state string as it 1) currently supports options that normal params don't and 2) could be useful for debugging?

mfix22: Yes i thought you were just going to 1. First check if `state` is in the query params, if so decode that and 2. Fallback to the rest of the query params if state is not there.

briandennis: Yeah, that was the plan 👍 Just thinking now being able to gen a state dump might be useful?

SirWindfield:
@briandennis how does the new api endpoint work? Is it just posting to `https://carbon.now.sh/image` and the body is a json holding all information in base64? I wanted to make a small cli for myself :smile:

PR #228, opened 2018-02-09
Add link to vim plugin.

kristijanhusak:
I created vim plugin https://github.com/kristijanhusak/vim-carbon-now-sh .


mfix22:
This is awesome! Thanks for doing this!

PR #237, opened 2018-02-15
Add support to keyboard navigation in Dropdowns

### LINKED ISSUE #154: Make it possible to change theme on arrow keypresses
Hi guys, thank you all for this awesome project! 👌🏼👏🏼 
_I've not used the issue template since this is a feature request_

First thing I tried to do on the web page was to change the theme on arrow keys to see how they look, similar to how VSCode does it. I noticed it wasn't possible and then tried to implement it. I have a somewhat working solution by now!

carbon

I would like to know if you guys think it's a good idea so that I can than create a PR to work better on it! 

I have yet to figure out how to make the dropdown also scroll (maybe keep a ref to it and use [scrollTop](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop) ? 🤔). Also, I'm not very familiar with Next.js... Is it ok to add and eventHandler to the window object directly?

Thank you all again! Haven't had the opportunity to use it but will definitely do it.

bruno02221:
## Description

It was added support to keyboard navigation in the Dropdown component using the arrow down and arrow up keys, so the user can navigate and have real-time feedback from the Themes and Languages using the keyboard. So it closes #154 

The Dropdown component was completely reimplemented using the [`downshift`](https://github.com/paypal/downshift) library (as suggested by @mfix22) because it makes the keyboard events handling easier, and it can make future implementation of features like `typeahead` easy too.


briandennis:
Thanks @bruno02221! This is exactly what we wanted 👍

mfix22:
@bruno02221 this is incredible. Thanks!!! Downshift worked great @kentcdodds

kentcdodds:
Niiiiice!

bruno02221:
👍

tibuurcio:
@bruno02221 Awesome work bruno! :+1: Sorry for the lack of response on the issue page, I got caught up with work.. 

Cheers from Brazil :)

Issue #241, opened 2018-02-16
Square background


besasam:
It would be great if there was a setting to give the editor a square background, for sharing on Instagram.
Could probably be done by adjusting the paddings upon resizing so that a square aspect ratio is maintained.

~~Tried doing it myself but I suck at React so I gave up lol~~

molnarmark:
I've just implemented this, if @briandennis or any other core maintainer thinks its a good idea to have this as an option, I'll open a PR.

briandennis:
Sounds good to me! @molnarmark your solution looks good, though we'd probably want this in the advanced section of the menu. Would happily accept a PR for this 👍

molnarmark:
#250 Opened.

PR #250, opened 2018-02-20
Feature: Squared Image

molnarmark:
The feature from #241, moved into the Advanced dropdown as requested.


jakedex:
This is great! Thanks for the contribution, @molnarmark 👍

Issue #254, opened 2018-02-21
Add Title


BLamy:
It would be nice if there was an ability to add a title to your code similar to this.

![image](https://user-images.githubusercontent.com/324928/36494156-47b6226c-16ff-11e8-996f-a78713ef0523.png)

molnarmark:
I'd be happy to implement this if needed!

KINGSABRI:
A title, or file name

jakedex:
I really like this idea! If you could implement @molnarmark, that'd be awesome. Maybe make the font-size a bit smaller, like the example below. 

![example](https://user-images.githubusercontent.com/10369094/36510867-27d79612-172a-11e8-9ebb-86df8cdd8617.png)

molnarmark:
What would be a good way to do this? Have an invisible input box at the top? I was thinking about adding it to the settings dropdown but it really doesn't belong there.

jakedex:
Yeah, let's throw an invisible input box at the top for now. We'll have to add some more UX cues down the road.

molnarmark:
I'm gonna add this and submit a PR soon in the next couple of days. 👍

molnarmark:
#257 was merged. 😄

mallamzamaani1:
This is Mallam I am a powerful spritual man who help people  u can whatsapp me now or call me now (+233546805897) ,for help like, 
1.Schooling
2.Riches
3.Instant money 
4.business improvement 
5.Child birth
6.Lotto numbers 
7.Scam powers 
8.protection 
All problems can be solve

PR #257, opened 2018-02-22
Feature: Window Title

molnarmark:
This PR is related to #254.

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -27,6 +27,7 @@ class Carbon extends React.Component {
     }
 
     this.handleLanguageChange = this.handleLanguageChange.bind(this)
+    this.handleTitleBarChange = this.handleTitleBarChange.bind(this)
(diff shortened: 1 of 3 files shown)


briandennis:
@molnarmark can you update this so that the title text is saved to the editor state object?

molnarmark:
Sure! Latest commit does it.

jakedex:
Looks great! Thanks for tackling this, @molnarmark 🎊

metaskills:
I've noticed that I can not get the window title to export using PNG, but it does work with SVG. Is that intentional?

mfix22:
Hey @metaskills can you post a link to an example where it is failing? I just tried it and it is working for me: 
carbon

metaskills:
Thanks for following up, after some testing, I found it does not work in Safari. Work fine in Chrome.

PR #261, opened 2018-02-23
Added missing settings to mapping in routing.js so they will update the url

### LINKED ISSUE #259: Some settings aren't exported to url
Setting font size to max

screenshot 2

After refreshing the page font size returns to default

screenshot 3

Other settings in this menu (line numbers for example) seem to be exported correctly, while others don't (font family for example).

Settings that are not exported correctly:

- Window theme
- Font family
- Font size
- Square image
- Drop shadow Y offset
- Drop shadow blur radius

DeepSpace2:
Closes #259

### DIFF
--- a/lib/routing.js
+++ b/lib/routing.js
@@ -20,13 +20,19 @@ const mapper = new Morph({
 const mappings = [
   { field: 'bg:backgroundColor' },
   { field: 't:theme' },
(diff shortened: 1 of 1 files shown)


mfix22:
Thanks so much for the help!

PR #278, opened 2018-03-08
Unsplash Integration

### LINKED ISSUE #269: Unsplash integration
After seeing @_developit 's beautiful tweet

image

I realized it'd be nice to have a button here to source a random CC-licensed image from Unsplash

image

jakedex:
Closes #269 🎊


components/RandomImage.js
```diff
@@ -0,0 +1,111 @@
...
+const RAND_RANGE = 1000000
+const UPDATE_SIZE = 20
+const WALLPAPER_COLLECTION_ID = 136026
+const RANDOM_WALLPAPER_URL = `https://source.unsplash.com/collection/${WALLPAPER_COLLECTION_ID}/240x320`
+
+const largerImage = url => url.replace(/w=\d+/, 'w=1920').replace(/&h=\d+/, '')
+const smallerImage = url => url.replace(/w=\d+/, 'w=240')
+
+const imageUrls = {}
+let cache = []
```
mfix22: @jakedex one more thing, and then i'll approve. I think the cache and the `imageUrl` should be instance variables, meaning it should be `this.cache` and `this.imageUrls`. That way each instance of this has its own cache, etc. I know we only have one instance but conventionally i think that's a good choice.

PR #293, opened 2018-03-21
Lint and fix issues, don't add linting yet though

mfix22:
~Wait for #287 to close first~


lib/api.js
```diff
@@ -41,10 +41,10 @@ const getGist = uid => {
 
 // private
 function openTwitterUrl(twitterUrl) {
-  const width = 575,
-    height = 400
+  const width = 575
+  const height = 400
   const left = (window.outerWidth - width) / 2
-  const right = (window.outerHeight - height) / 2
+  const top = (window.outerHeight - height) / 2
```
mfix22: Hey @jakedex can you give this part a look quick? There was no `top` before, and `right` wasn't being used so i was assuming those just needed to be switched.

PR #301, opened 2018-03-28
Use Unsplash API

jakedex:
TODO

- [ ] Add photographer name to carbon image
- [ ] Fix error when setting background image (`Warning: Can only update a mounted or mounting component.`)
- [ ] Clean up server imports/handler exports
- [ ] Add [link to Unsplash](https://medium.com/unsplash/unsplash-api-guidelines-28e0216e6daa) to image thumbnail
- [ ] Submit api [verification screenshots](https://unsplash.com/documentation#registering-your-application)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "graphql": "^0.11.7",
     "highlight.js": "^9.12.0",
     "history": "^4.7.2",
+    "isomorphic-fetch": "^2.2.1",
     "lodash.debounce": "^4.0.8",
     "match-sorter": "^2.2.0",
     "morgan": "^1.8.2",
@@ -53,7 +54,8 @@
     "react-syntax-highlight": "^15.3.1",
     "resize-observer-polyfill": "^1.5.0",
     "tohash": "^1.0.2",
-    "twitter": "^1.7.1"
+    "twitter": "^1.7.1",
+    "unsplash-js": "^4.8.0"
   },
   "devDependencies": {
     "@zeit/next-css": "^0.1.5",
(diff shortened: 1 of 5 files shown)


mfix22:
What do you all think about clicking the thumbnail image to use it and having the "Use Image" link be the link to the "Source" instead?

briandennis:
I like the idea. I think we should add some UI cue that clicking the image will use it in that case. Maybe dim on hover and pop  up the text _Set Background_  or something?

Instead of linking to _Source_ wanna link to _Image by Author Name_? Regardless, we should probably also add the author attribution link to the image crop screen after it's selected as well.

Happy to build these out if you agree

mfix22:
@briandennis i dig that, go for it. Let's build it out on this branch though, since I'd like to wait until our API request rate increase is accepted before merging 👍

package.json
```diff
@@ -76,7 +76,8 @@
       "TWITTER_ACCESS_TOKEN_SECRET": "@twitter-access-token-secret",
       "LOGS_SECRET_PREFIX": "@logs_secret_prefix",
       "UNSPLASH_SECRET_KEY": "@unsplash_secret_key",
-      "UNSPLASH_ACCESS_KEY": "@unsplash_access_key"
+      "UNSPLASH_ACCESS_KEY": "@unsplash_access_key",
+      "UNSPLASH_CALLBACK_URL": "@unsplash_callback_url"
```
jakedex: Wait, why was this added?

mfix22:
@jakedex the client wasn’t getting instantiated correctly for me locally
without that param.

mfix22:
Final TODO before merging: 
- [ ] Confirm API rate limit increase

PR #309, opened 2018-04-01
improve first meaningful paint on browser

IanWang:
### Before (~2s)
screen shot 2018-04-02 at 00 41 58

### After (~1s)
screen shot 2018-04-02 at 00 41 10

☝️ both are tested on localhost (Chrome v65.0)
the green line is the first frame captured.

## Description
speed up by delaying codemirror theme & highlightjs (big lib)

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -1,4 +1,3 @@
-import * as hljs from 'highlight.js'
 import React, { PureComponent } from 'react'
 import Spinner from 'react-spinner'
 import ResizeObserver from 'resize-observer-polyfill'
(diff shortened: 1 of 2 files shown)


mfix22:
Hey @IanWang this is great! As for checking if rendering if occurring inBrowser, are you sure this isn't causing a rerender due to a client-server mismatch (should have a warning in the console).

A way to avoid this is to just set `state.mounted: true` after `componentDidMount()` and render based on `this.state.mounted === true`

IanWang:
@mfix22 yes I was expecting that, but interestingly I didn't see any warning about the mismatch.
maybe it's because React 16 changes the algo? or it's because of the implementation of `next/head`.

mfix22:
Interesting. Okay well if i see it later i'll know where to look! 😄 Thanks so much.

Issue #307, opened 2018-03-31
Refactor state 


mfix22:
- [x] `Editor` should be a component, rather than a page
- [x] State should be a container that is separate from any styles/templates
  - https://github.com/jamiebuilds/unstated

PR #310, opened 2018-04-02
Refactor index.js

mfix22:
Start on #307

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -1,4 +1,5 @@
 import React, { PureComponent } from 'react'
+import * as hljs from 'highlight.js'
 import Spinner from 'react-spinner'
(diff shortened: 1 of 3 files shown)


components/Carbon.js
```diff
@@ -50,20 +51,13 @@ class Carbon extends PureComponent {
     this.props.updateTitleBar(newTitle)
   }
 
-  async getHighlightLang(code = '') {
```
jakedex: Was this not working?

mfix22: It was just such a critical action (highlighting the code based on language when the user lands on the page) that i didn't want to delay it with this extra dynamic import

mfix22:
@jakedex i am just going to update carbon-editor to be up to date with this current editor. We can even maybe publish it from this repo. We will see.

PR #313, opened 2018-04-06
Add Fantasque Sans Mono font

camflan:
This patch adds the Fantasque Sans Mono font as an option.

The font is open source here,
https://github.com/belluzj/fantasque-sans


mfix22:
Thanks! 😄

PR #319, opened 2018-04-08
Adds Atom's One Dark theme

bjervisSEEK:
## New Theme
This PR is to add Atom's One Dark theme to Carbon

## Description
The Codemirror compliant CSS is from [Aerobird98](http://github.com/Aerobird98)'s codemirror-one-dark-theme npm package. PR adds the CSS file to customModes and adds the constant to constants.js


mfix22:
Hey @bjervisSEEK  all CSS files must be placed in `/static` (see react-spinner.css) and imported in `Meta.js`, or placed as a `<style>` tag in Meta.

mfix22:
@bjervisSEEK if would be great if you could place this as a CSS file under `/static/customThemes/one-dark.css`

bjervisSEEK:
Thanks @mfix22, I've just pushed an update which hopefully puts it in the right spot.

components/Meta.js
```diff
@@ -34,6 +34,7 @@ export default () => {
         <meta name="theme-color" content="#121212" />
         <title>Carbon</title>
         <link rel="shortcut icon" href="/static/favicon.ico" />
+        <link rel="stylesheet" href="/static/customThemes/one-dark.css" />
```
mfix22: sorry to make this so complicated, we will try to document this process better in the future. Can you actual just do `import '../static/customThemes/one-dark.css'` like the others at the top of this file?

lib/constants.js
```diff
@@ -56,6 +56,11 @@ export const THEMES = [
     id: 'oceanic-next',
     name: 'Oceanic Next'
   },
+  {
+    id: 'one-dark',
+    name: 'One Dark',
+    link: 'customModes/onedark.css'
```
mfix22: Can you remove this `link` attribute? It's not being used 😄

bjervisSEEK:
Thanks @mfix22. It didn't look like anyone has added a custom theme before, so didn't have a lot to go on. Hopefully these two commits should address those review items.

Issue #321, opened 2018-04-10
localStorage config is ignored when url contains only language and code


stevensacks:
I'm using 3rd party plugins in my various IDEs and they all use the above url with just language and code and they all end up causing this bug.

## Expected Behavior
http://carbon.now.sh/?l=javascript&code=const%20foo%20%3D%20%27bar%27%3B

Should open Carbon with my localStorage config for styling filled in.

## Actual Behavior
Opens the code with Carbon's default config instead of the saved config, and, as a result, overwrites the localStorage config with Carbon's default, meaning I have to manually set all my config back again.

mfix22:
Currently if you pass a query param, localStorage will be ignored. That is operating as it is intended. 

Are you suggesting that you want parts to be loaded from that query param and others from `localStorage`?

stevensacks:
IDE plug-ins that allow you to select code and then open in Carbon (very convenient) will only pass language and code in the URL.

I think that if you _only_ pass language and code then you should look at localStorage for the rest.

stevensacks:
I was looking through the code to try and find the place that handles this to make a pull request, but I haven’t been able to find it yet.

mfix22:
Although we don't want to override fields that were set by URL with `localStorage`. In your example, those fields would be overridden. 

You have to calculate the difference between `urlConfig` and `localStorageConfig` and pass that to `setState`

stevensacks:
So that's just an order change

```javascript
{
    ...defaultConfig,
    ...localStorageConfig,
    ...urlConfig,
}
```

stevensacks:
It could be as simple as this

```
    this.state = Object.assign(
      {
        ...DEFAULT_SETTINGS,
        uploading: false,
        code: props.content,
        ...getState(localStorage),
      },
      this.props.initialState
    )
```

And erase componentDidMount.

mfix22:
You can't access localStorage in `getInitialProps()`

stevensacks:
Yeah I figured it out. Really simple fix. Check my fix.

stevensacks:
I'm not sure if there are any issues from delaying the initialState being set until mount, but it seems ok based on what I can see.

stevensacks:
Submitted a PR.

PR #322, opened 2018-04-10
fix issue #321 - load defaults, then localStorage, then url config

stevensacks:
## Description
For IDEs that allow you to open selected code in Carbon, they only pass the language and the code. To avoid losing your localStorage config, the settings should be merged in the following order:

1. DEFAULT
2. localStorage
3. urlParams

This ensures that all required config is present and that url params trump all.

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -62,9 +62,8 @@ class Editor extends React.Component {
   }
 
   componentDidMount() {
-    // Load from localStorage instead of query params
+    // Load from localStorage and then URL params
     this.setState({
-      ...this.state,
       ...getState(localStorage),
       ...this.props.initialState
     })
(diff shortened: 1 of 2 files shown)


components/Editor.js
```diff
@@ -66,12 +63,11 @@ class Editor extends React.Component {
...
   componentDidMount() {
     // Load from localStorage instead of query params
-    if (!this.props.initialState) {
-      const state = getState(localStorage)
-      if (state) {
-        this.setState(state)
-      }
-    }
+    this.setState({
+      ...this.state,
```
mfix22: You don't need to include `this.state` in `setState` (:

stevensacks: Fixed

mfix22:
@stevensacks thanks so much no idea why I didn't write it like this the first time. _Way_ better. 

@briandennis see any issues?

stevensacks:
Second set of eyes always helps. :)

Great tool! Happy to help make it just a little bit better.

stevensacks:
Let me know when it's deployed!

Issue #275, opened 2018-03-07
Some sort of runtime and/or static type checking 


wuweiweiwu:
via `prop-types` and/or `flow`

I think this would make development a lot easier :)

mfix22:
@wuweiweiwu `prop-types` would definitely be beneficial, but we don't want to add TypeScript or Flow at this point.

Are you interested in submitting a PR for this?

wuweiweiwu:
@mfix22 Yep! I'll get started on that:)

wuweiweiwu:
@mfix22 sounds good! Thanks for letting me know!

mfix22:
At this state, with component APIs in flux, we want to avoid prop-types for now. We did just add legit linting though, so I hope that helps

PR #338, opened 2018-04-19
Refactor state components

### LINKED ISSUE #307: Refactor state

mfix22:
This is just the initial PR. In the future though we can: 
- [x] Separate editor into `Component` (contains container instance) and function (responsible for all DOM elements)
- [x] Remove prop-drilling and have deeper components subscript to editor state
  - [x] Make `<Toolbar/>` (Buttons and settings) `<TextInput>` (Carbon + DnD target) top level with no function props
- [ ] separate state into multiple containers
- [x] Editor -> Carbon, Carbon -> CodeWindow, CodeWindow-> Editor

- Closes #307 
Will reopen another issue with other TODOs


mfix22:
- Adds linting
  - Closes #275 
- Removes old dependencies

components/Coder.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
```
jakedex: `Coder` seems kinda confusing to me, how about `CarbonContainer` or something?

components/Coder.js
```diff
@@ -0,0 +1,55 @@
...
+import ReadFileDropContainer, { DATA_URL, TEXT } from 'dropperx'
+
+import EditorContainer from '../containers/Editor'
+import Overlay from './Overlay'
+import Carbon from './Carbon'
+
+import { DEFAULT_CODE } from '../lib/constants'
+import { isImage } from '../lib/util'
+
+const to = [EditorContainer]
```
jakedex: Does this need to be a variable? Might be more readable w/o it...

mfix22: I just want to make sure that the props is equivalent as to not cause rerenders do to new array constructions

components/Coder.js
mfix22: I want to rename our Carbon to Coder, the current Coder component to Editor, and then the Editor component to Carbon. What do you think?

components/Carbon.js
```diff
@@ -1,279 +1,41 @@
...
-      )
-    }
-
     return (
-      <div id="section">
-        <div id="export-container" ref={ele => (this.exportContainerNode = ele)}>
-          {content}
-          <div id="twitter-png-fix" />
+      <Provider inject={this.inject}>
+        <div id="editor">
```
jakedex: Let's use #carbon here

components/Editor.js
```diff
@@ -1,276 +1,48 @@
...
+        )}
+      </ReadFileDropContainer>
     )
   }
-}
 
-function isImage(file) {
-  return file.type.split('/')[0] === 'image'
+  render() {
+    return <Subscribe to={to}>{this.renderPane}</Subscribe>
```
briandennis: let's rename `to` to something more meaningful

components/CodeWindow.js
```diff
@@ -0,0 +1,280 @@
...
+  codeUpdated(newCode) {
+    this.handleLanguageChange(newCode)
+    this.props.updateCode(newCode)
+  }
+
+  handleTitleBarChange(newTitle) {
+    this.props.updateTitleBar(newTitle)
+  }
+
+  handleLanguageChange = debounce(
```
briandennis: Can this be extracted to the state container?

containers/Editor.js
```diff
@@ -0,0 +1,168 @@
...
+class EditorContainer extends Container {
+  constructor(props = {}) {
+    super(props)
+    this.state = {
+      ...DEFAULT_SETTINGS,
+      uploading: false,
+      code: props.content
+    }
+
+    this.save = this.save.bind(this)
```
briandennis: are any of these private? If so, maybe specify which are

components/CodeWindow.js
mfix22: Yes. But can we extract it later? there are a bunch of spots we can add new subscriptions and containers, but the goal of this PR is mostly to refactor the top level design and let Carbon be exportable on it's own.

containers/Editor.js
mfix22: None of them are as of now. Just rename one last one

PR #369, opened 2018-05-20
Add Octave language support

anku255:
I have added Octave language support. Since Octave and Matlab follow the same syntax hence for the name I chose `Octave/Matlab`. If you don't like that then I can rename it to `Octave`. 

It looks like below:
image

I am a newbie when it comes to contributing to open source so please bear with me if I broke something. I wanted to test before submitting but I failed to get Cypress working on my windows machine. Also, I have no idea how README.md got changed (maybe it changed in pre-commit  step). Please take a look and provide feeback.


lib/constants.js
```diff
@@ -310,6 +310,12 @@ export const LANGUAGES = [
     mime: 'text/x-objectivec',
     short: 'objectivec'
   },
+  {
+    name: 'Octave/Matlab',
```
mfix22: This looks great! Can you actually name it `"MATLAB/Octave"`?

anku255:
Thank you so much @mfix22  for accepting my PR. May I ask when the changes will be reflected on the site?

mfix22:
@anku255 no problem! I am going to try and get it deployed today.

PR #372, opened 2018-05-21
add alt text support for Twitter images

### LINKED ISSUE #361: Provide code as alt text when uploading to Twitter
Relevant section of Twitter API docs:
https://developer.twitter.com/en/docs/media/upload-media/api-reference/post-media-metadata-create

briandennis:
closes #361

### DIFF
--- a/handlers/twitter.js
+++ b/handlers/twitter.js
@@ -11,6 +11,10 @@ const client = new Twitter({
   access_token_secret: process.env.TWITTER_ACCESS_TOKEN_SECRET
 })
 
+const extractMediaId = morph.get('data.media_id_string')
+const extractImageUrl = morph.get('data.entities.media.0.display_url')
+const extractErrorCode = morph.get('0.code')
+
 const uploadImage = data => client.post('media/upload', { media_data: data })
 const uploadMetadata = (altText, twitterRes = {}) => {
   if (!altText)
@@ -31,10 +35,6 @@ const uploadTweet = (twitterRes = {}) => client.post('statuses/update', {
     status: `Carbon Copy #${extractMediaId(twitterRes).slice(0, 8)}`,
     media_ids: extractMediaId(twitterRes)
 })
-
-const extractMediaId = morph.get('data.media_id_string')
-const extractImageUrl = morph.get('data.entities.media.0.display_url')
-const extractErrorCode = morph.get('0.code')
 
 const respondSuccess = (res, url) => res.json({ url })
 const respondFail = (res, err) => {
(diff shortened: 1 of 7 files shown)


mfix22:
This is unrelated to review, but @briandennis can you fix the linting?

handlers/twitter.js
```diff
@@ -1,23 +1,39 @@
-const Twitter = require('twitter')
+const Twitter = require('twit')
 const morph = require('morphmorph')
 
 const RATE_LIMIT_CODE = 420
+const MAX_ALT_TEXT_LENGTH = 420
```
mfix22: Funny these two values are the same . . .

handlers/twitter.js
```diff
@@ -1,23 +1,39 @@
...
-  client.post('statuses/update', {
-    status: `Carbon Copy #${media.media_id_string.slice(0, 8)}`,
-    media_ids: media.media_id_string
-  })
+const uploadMetadata = (altText, twitterRes = {}) => {
+  if (!altText)
+    return twitterRes
 
-const extractImageUrl = morph.get('entities.media.0.display_url')
+  const formattedAltText = altText.length > MAX_ALT_TEXT_LENGTH
```
mfix22: Just FYI (and maybe we want to make this change) but if the length is shorter than the second `slice` param if will just return the whole string. So you don't need the conditional check. 

Are you making the check because you wanted it `MAX_ALT_TEXT_LENGTH - 3`? Because we could probably just do `.slice(0, MAX_ALT_TEXT_LENGTH - 3)` no matter.

handlers/twitter.js
```diff
@@ -1,23 +1,39 @@
...
+      alt_text: { text: formattedAltText }
+    })
+    .then(() => twitterRes)
+}
+const uploadTweet = (twitterRes = {}) => client.post('statuses/update', {
+    status: `Carbon Copy #${extractMediaId(twitterRes).slice(0, 8)}`,
+    media_ids: extractMediaId(twitterRes)
+})
+
+const extractMediaId = morph.get('data.media_id_string')
```
mfix22: Should we move these constants above where they are used? It works, but.

handlers/twitter.js
briandennis: I cannot imagine it was happenstance

handlers/twitter.js
briandennis: Yup, but you'd still need a length check for the ellipsis

handlers/twitter.js
briandennis: updated 👍

PR #384, opened 2018-06-02
Line number seperator

### LINKED ISSUE #351: Add line number vertical line
## Actual Behavior

carbon

Complaint: The space between line number and the actual code is too close. On some themes, the color similarity of line number and the code is confusing.

## Expected Behavior

carbon

Suggestions: 
1. The space between line number and the code should be a little bit wider.
2. Add a vertical line to clearly distinguish line number and the code.

adrianhartanto0:
## Description
To close #351. Added an optional setting option to show/hide a vertical line separator between the line numbers and the source code content.


components/Carbon.js
```diff
@@ -22,6 +22,20 @@ class Carbon extends PureComponent {
     this.handleLanguageChange = this.handleLanguageChange.bind(this)
     this.handleTitleBarChange = this.handleTitleBarChange.bind(this)
     this.codeUpdated = this.codeUpdated.bind(this)
+    this.codeMirrorUpdate = this.codeMirrorUpdate.bind(this)
+  }
+
+  codeMirrorUpdate(){
+    if(this.props.config.lineNumberSeperator){
```
mfix22: I know i mentioned calculating the color, but this seems like a lot of overhead for a small feature. Is there a color that works well with all the themes? Like `#fff`?

Sorry for making you do this

components/Carbon.js
```diff
@@ -118,13 +137,19 @@ class Carbon extends PureComponent {
...
             options={options}
+            onUpdate={this.codeMirrorUpdate}
           />
           {config.watermark && <Watermark />}
           <div id="container-bg">
             <div className="white eliminateOnRender" />
             <div className="alpha eliminateOnRender" />
             <div className="bg" />
           </div>
+          <style>
```
mfix22: You can just put this style in the other `style` tag

adrianhartanto0: Do you mean the `style` tag that begins with `#container`? I tried, the style will not be applied

mfix22: That is just because your classes are created outside of react. You can just add this class and then dynamically apply styles within the template string something like:
```js
this.props.config.lineNumberSeperator  ? `border-right: ${"1px solid white"};` : ''
this.props.config.lineNumberSeperator  ? `padding-right: 10px important;` : ''
```

and then remove the `codeMirrorUpdate` altogether.

mfix22:
Closing as stale. Please reopen if you have changes 👍

PR #390, opened 2018-06-07
Change a ? b : null to a && b in JSX

coolreader18:
## Description
Same as in the title, React accepts `false` as a child, so there is no need to do `cond ? <Elem /> : null` in JSX, you can just do `cond && <Elem />`.

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -109,13 +109,13 @@ class Carbon extends PureComponent {
     if (this.state.loading === false) {
       content = (
         <div id="container">
-          {config.windowControls ? (
+          {config.windowControls && (
             <WindowControls
               titleBar={this.props.titleBar}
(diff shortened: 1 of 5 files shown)


mfix22:
@coolreader18 this is definitely useful, but for certain cases a ternary is better (like `0 && <div/>` will render a `0` on the screen and we don't want that). 

I'm not sure if we should make the change because after getting bit by not having a ternary a few times now, I am starting to be in the "ternary-always- camp. 

@briandennis what do you think (as we form our style guide)?

briandennis:
@mfix22 I agree with you about the ternary. I think it's also more explicit about what it's doing in this case (switching between two values based on a condition). The `&&` solution also relies on the reader understanding in-depth the way logical operators return original values and short circuit.

Issue #395, opened 2018-06-12
How change watermark?


hamedmoody:
HI
How can i change watermark?

kaicianflone:
@mfix22 I have noticed you are using `dropperx` I will go ahead and implement custom watermark upload using it. I might have some questions though :)

mfix22:
Yeah go for it! I am pretty sure it will be fairly easy, but we might need to adjust the DropTarget box because the Watermark one will be _inside_ the other . . .

mfix22:
Closing this since the question was answered!

PR #397, opened 2018-06-15
Added the ability to add a custom watermark

kaicianflone:
Added the ability to upload a custom watermark from issue #395 

Currently only square watermarks will work and I have them scaled down to 75px * 75px.

Currently your watermark will also save as the watermark setting until you press reset default settings too 👍

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -106,6 +106,18 @@ class Carbon extends PureComponent {
         </style>
       </div>
     )
+    let watermark = <Watermark />
+    if (this.props.customWatermark) {
+      let watermarkSource = this.props.customWatermark
+      watermark = (
+        <img
+          className="watermark"
+          style={{ height: '75px', width: '75px' }}
+          src={watermarkSource.customWatermark}
+          alt="Custom Watermark"
+        />
+      )
+    }
     if (this.state.loading === false) {
       content = (
         <div id="container">
@@ -122,7 +134,7 @@ class Carbon extends PureComponent {
             value={this.props.children}
             options={options}
           />
-          {config.watermark && <Watermark />}
+          {config.watermark && watermark}
           <div id="container-bg">
             <div className="white eliminateOnRender" />
             <div className="alpha eliminateOnRender" />
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -107,13 +107,12 @@ class Carbon extends PureComponent {
       </div>
     )
     let watermark = <Watermark />
-    if (this.props.customWatermark) {
-      let watermarkSource = this.props.customWatermark
+    if (this.props.config.customWatermark) {
       watermark = (
         <img
           className="watermark"
           style={{ height: '75px', width: '75px' }}
-          src={watermarkSource.customWatermark}
+          src={this.props.config.customWatermark}
           alt="Custom Watermark"
         />
       )
--- a/components/CustomWatermarkPicker.js
+++ b/components/CustomWatermarkPicker.js
@@ -10,18 +10,14 @@ export default class extends React.Component {
   selectImage(e) {
     const file = e.target ? e.target.files[0] : e
 
-    return fileToDataURL(file).then(customWatermark => this.props.onChange(customWatermark))
+    return fileToDataURL(file).then(this.props.onChange)
   }
 
   render() {
     return (
       <div className={`toggle`}>
         <span>
-          <input
-            type="file"
-            accept="image/x-png,image/jpeg,image/jpg,image/svg+xml"
-            onChange={this.selectImage}
-          />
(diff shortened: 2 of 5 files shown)


components/Settings.js
```diff
@@ -105,6 +106,7 @@ class Settings extends React.Component {
               enabled={this.props.watermark}
               onChange={this.props.onChange.bind(null, 'watermark')}
             />
+            {this.props.watermark && <CustomWatermarkPicker onChange={this.props.onUpdate} />}
```
mfix22: Just use `this.props. onChange` like the other settings do.

components/Carbon.js
```diff
@@ -106,6 +106,18 @@ class Carbon extends PureComponent {
...
       </div>
     )
+    let watermark = <Watermark />
+    if (this.props.customWatermark) {
+      let watermarkSource = this.props.customWatermark
+      watermark = (
+        <img
+          className="watermark"
+          style={{ height: '75px', width: '75px' }}
+          src={watermarkSource.customWatermark}
```
mfix22: You have `this.props.customWatermark.customWatermark.` . . . are you able to flatten this? (:

components/CustomWatermarkPicker.js
```diff
@@ -0,0 +1,59 @@
...
+  constructor(props) {
+    super(props)
+    this.selectImage = this.selectImage.bind(this)
+  }
+
+  selectImage(e) {
+    const file = e.target ? e.target.files[0] : e
+
+    return fileToDataURL(file).then(dataURL =>
+      this.setState(() => {
```
mfix22: Why are you calling `setState` here? You don't need to if you don't have any state.

lib/constants.js
```diff
@@ -487,5 +487,6 @@ export const DEFAULT_SETTINGS = {
   exportSize: '2x',
   titleBar: '',
   watermark: false,
+  customWatermark: false,
```
mfix22: I would make this `null` since it doesn't become a boolean property.

components/CustomWatermarkPicker.js
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className={`toggle`}>
+        <span>
+          <input type="file" onChange={this.selectImage} />{' '}
+          <span className="labelText"> Upload Custom Watermark </span>
+        </span>
+
+        <style jsx>
+          {`
+            .toggle {
```
mfix22: @jakedex  mind reviewing this for style changes whenever you get a sec?

kaicianflone:
Ok thank you for the review @mfix22 :) That was my first PR and a very good learning experience. I am going to add some testing too. Thanks!

components/CustomWatermarkPicker.js
```diff
@@ -10,20 +10,18 @@ export default class extends React.Component {
...
   }
 
   render() {
     return (
       <div className={`toggle`}>
         <span>
-          <input type="file" onChange={this.selectImage} />{' '}
+          <input
+            type="file"
+            accept="image/x-png,image/jpeg,image/jpg,image/svg+xml"
```
mfix22: `image/*` (:

components/CustomWatermarkPicker.js
```diff
@@ -10,20 +10,18 @@ export default class extends React.Component {
...
     const file = e.target ? e.target.files[0] : e
 
-    return fileToDataURL(file).then(dataURL =>
-      this.setState(() => {
-        this.props.onChange({
-          customWatermark: dataURL
-        })
-      })
-    )
+    return fileToDataURL(file).then(customWatermark => this.props.onChange(customWatermark))
```
mfix22: ` fileToDataURL(file).then(this.props.onChange)`

mfix22:
@kaicianflone this is looking a lot better! These merge conflicts ^^ should be super easy to fix (sorry about that). Also what do you think about just making the watermark itself the input you click on to change the image? That way we don't need to crowd the settings and it is more natural to change the image?

kaicianflone:
Oh yeah I like that idea. It should be very smooth. I will go ahead and make the change :)

Also I wanted to ask if I should style the Height/Width to 75 x 75 px or if I should just set the maxHeight / maxWidth to 75px or 100px perhaps if anyone wanted to do a 4:3 watermark instead of the 1:1. What do you think?

mfix22:
@kaicianflone  i think you should do a max-width and max-height yeah 👍

mfix22:
hey @kaicianflone 🙂where are you at with this PR?

mfix22:
Hey @kaicianflone we are really excited about this feature. Are you able to update it based on what we discussed above? 🙂

mfix22:
Closing as stale. Please reopen if you have changes 👍

Issue #406, opened 2018-06-22
Adjust line height


jakearchibald:
It'd be great to be able to adjust the line height. I fine code a lot easier to read with a little more line spacing.

jakearchibald:
Fwiw, you can adjust the line height with devtools & Carbon keeps it for the screenshot.

Here's `line-height: 1.6`:

![Example](https://user-images.githubusercontent.com/93594/41770303-aa5eed04-7609-11e8-951b-cd0b482736ed.png)

mfix22:
@jakearchibald my initial thought is that this is an advanced use case that most people won't use, and could potentially lead to a "feature-spiral" where we have to add one setting for every CSS property. 

I am going to consult with the rest of the core team before closing, though.

jakearchibald:
That's fair. I guess I could apply it as a custom stylesheet.

If you do want the feature, I'd be happy to create a PR.

mfix22:
@jakearchibald talked with the team today, and we would be happy to support `line-height`, as I found out that is a common property to adjust in a lot of web tools 😄 

Would _love_ to accept a PR 👍

PR #410, opened 2018-06-26
Adding line-height setting & granular font size.

jakearchibald:
Added line-height setting as per https://github.com/dawnlabs/carbon/issues/406.

While I was there, I also changed font-size so it increments by 0.5, since the steps were pretty big.

### DIFF
--- a/components/Settings.js
+++ b/components/Settings.js
@@ -51,6 +51,7 @@ class Settings extends React.Component {
             value={this.props.fontSize || 13}
             minValue={10}
             maxValue={18}
+            step={0.5}
             onChange={this.props.onChange.bind(null, 'fontSize')}
           />
           <Toggle
--- a/components/Settings.js
+++ b/components/Settings.js
@@ -70,6 +70,14 @@ class Settings extends React.Component {
           />
           <Collapse label="Advanced">
             <Slider
+              label="Line height"
+              value={this.props.lineHeight}
+              minValue={90}
+              maxValue={250}
+              usePercentage={true}
+              onChange={this.props.onChange.bind(null, 'lineHeight')}
+            />
+            <Slider
               label="Padding (vertical)"
               value={this.props.paddingVertical || 16}
               maxValue={200}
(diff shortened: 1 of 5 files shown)


lib/routing.js
```diff
@@ -32,6 +32,7 @@ const mappings = [
   { field: 'ln:lineNumbers', type: 'bool' },
   { field: 'fm:fontFamily' },
   { field: 'fs:fontSize' },
+  { field: 'lh:lineHeight' },
```
jakearchibald: I'm cargo-culting here. Is this right?

lib/constants.js
```diff
@@ -481,6 +481,7 @@ export const DEFAULT_SETTINGS = {
   language: DEFAULT_LANGUAGE,
   fontFamily: 'Hack',
   fontSize: '14px',
+  lineHeight: '133%',
```
jakearchibald: The app currently uses `4 / 3` as the line height. This is the closest value that's an integer.

components/Slider.js
```diff
@@ -13,6 +13,7 @@ export default class extends React.Component {
   render() {
     const minValue = this.props.minValue || 0
     const maxValue = this.props.maxValue || 100
+    const step = 'step' in this.props ? this.props.step : 1;
```
jakearchibald: 0 is a valid step value, so a simple truthy check isn't good enough.

components/Settings.js
```diff
@@ -69,6 +70,14 @@ class Settings extends React.Component {
...
           />
           <Collapse label="Advanced">
+            <Slider
+              label="Line height"
+              value={this.props.lineHeight}
+              minValue={90}
+              maxValue={250}
+              usePercentage={true}
+              onChange={this.props.onChange.bind(null, 'lineHeight')}
+            />
```
jakearchibald: Happy to move this, but I put it to the top so it's closest to the font-size control while still being in the 'advanced' settings.

jakedex:
This is great! Thanks, @jakearchibald 👍

PR #412, opened 2018-06-27
fix base64 bug

mashihua:
The Base64 table has Char '+' [https://en.wikipedia.org/wiki/Base64](https://en.wikipedia.org/wiki/Base64), it will turn to blank characters in url request, It will cause some error when  deserializeState the state.

### DIFF
--- a/handlers/image.js
+++ b/handlers/image.js
@@ -9,7 +9,7 @@ module.exports = browser => async (req, res) => {
   if (!state) res.status(400).send()
 
   try {
-    await page.goto(`http://localhost:${PORT}?state=${state}`)
+    await page.goto(`http://localhost:${PORT}?state=${encodeURIComponent(state)}`)
     await page.addScriptTag({ path: './lib/customDomToImage.js' })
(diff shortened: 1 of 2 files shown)


mfix22:
Hey @mashihua thanks for the PR!

handlers/image.js
```diff
@@ -9,7 +9,7 @@ module.exports = browser => async (req, res) => {
   if (!state) res.status(400).send()
 
   try {
-    await page.goto(`http://localhost:${PORT}?state=${state}`)
+    await page.goto(`http://localhost:${PORT}?state=${encodeURIComponent(state)}`)
```
briandennis: We should be handling this properly in the [serialize/deserialize functions](https://github.com/dawnlabs/carbon/blob/master/lib/routing.js#L50) rather than inline on the handler

mfix22: Hey @mashihua can you move `encodeURIComponent` into the function @briandennis mentioned above? (:

PR #418, opened 2018-07-01
Allow changing transparency and backdrop blur

astronomersiva:
Allows setting window transparency and blur.

Addresses #123.

screen shot 2018-07-01 at 3 23 30 pm

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -35,6 +35,8 @@ class Carbon extends PureComponent {
     this.codeUpdated = this.codeUpdated.bind(this)
     this.onBeforeChange = this.onBeforeChange.bind(this)
     this.handleLanguageChange = this.handleLanguageChange.bind(this)
+
+    this.codeMirrorNode = React.createRef();
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


components/Carbon.js
```diff
@@ -83,6 +85,29 @@ class Carbon extends PureComponent {
...
+        let alpha = parseFloat(transparency) / 100
+        if (backgroundColor.includes('rgba')) {
+          background = backgroundColor.split(',')
+          background[3] = ` ${alpha})`
+          background = background.join(', ')
+        } else {
+          background = backgroundColor.replace(')', `, ${alpha})`).replace('rgb(', 'rgba(')
+        }
+
+        background = `${background} !important`
```
astronomersiva: some themes have !important in their background

components/Carbon.js
```diff
@@ -83,6 +85,29 @@ class Carbon extends PureComponent {
         'Shift-Tab': 'indentLess'
       }
     }
+
+    let background = '';
+    if (!config.didThemeChange) {
```
astronomersiva: Used to keep track of whether a theme was changed...without this check, the background will not change between different themes

astronomersiva:
The exported PNG does not contain the blur effect but the exported SVG does..Pretty sure this is because of dom-to-image.

PR #419, opened 2018-07-03
Added sTeX/LaTeX Language Support

tirth14:
This PR adds the language support for sTeX/LaTeX.

## Description

The constant id for sTeX/LaTeX has been added which is supported in Codemirror.

This is how your latex code would look like in Carbon:

carbon 1

Thanks.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -274,6 +274,10 @@ export const LANGUAGES = [
     short: 'kotlin'
   },
   {
+    name: 'LaTeX',
+    mode: 'stex'
+  },
+  {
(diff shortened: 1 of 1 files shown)


lib/constants.js
```diff
@@ -367,6 +367,10 @@ export const LANGUAGES = [
     name: 'SQL',
     mode: 'sql'
   },
+  {
+    name: 'sTeX/LaTeX',
```
mfix22: Can you change it to just `LaTeX`? (:

tirth14: Sure 👍

PR #423, opened 2018-07-06
Fix crash on up/down press in Dropdown

### LINKED ISSUE #420: Up/Down keyboard
Hello.

image

It's happened when selected style or language and press up/down keyboard buttons.

samiskin:
Fixes #420 where the application crashes after selecting a dropdown item
and then pressing the up or down arrow keys.

## Description
The app would crash because the default highlighted index for the Dropdown was from `this.props.list` while the actual items being displayed were in `this.state.itemsToShow`.  When up/down was pressed this discrepancy would cause an error.

`this.userInputtedValue` also changed even on just clicking on an item rather than just manually typing the value.  Once the crash was fixed, this issue would mean the list would only show 1 item since it would think the user inputted the full word manually.

### DIFF
--- a/components/Dropdown.js
+++ b/components/Dropdown.js
@@ -16,9 +16,11 @@ class Dropdown extends PureComponent {
       if (Object.prototype.hasOwnProperty.call(changes, 'inputValue')) {
         if (changes.type === Downshift.stateChangeTypes.keyDownEscape) {
           inputValue = this.userInputtedValue
-        } else {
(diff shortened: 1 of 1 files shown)


samiskin:
When I tried using either `npm` or `yarn` to install this project's dependencies, it would error on installing `cypress`, so I couldn't actually run the tests 😢

mfix22:
Thanks so much @samiskin 👍

PR #430, opened 2018-07-13
Add optional timestamp to saved file names

erickzhao:
## Motivation

I was using Carbon this week (great tool btw), and thought to myself that the user experience might be a bit better if the file names were timestamped, like how the native screen capture tool on macOS saves files as `Screen Shot 2018-07-13 at 1.47.37 AM` or how [MonoSnap](https://monosnap.com/welcome) saves files as `screencast 2018-07-12 14-24-21.mp4`

It's a minor improvement, but I personally find it easier to sort through multiple Carbon-generated files if they all have unique file names like this, since:
* Date/time is a lot easier to read through and gives more information than `carbon, carbon(1), carbon(2), etc.`.
* The order in which your files are displayed is consistent with time. If you instead save all files with the same name, you can get mixed up while deleting (e.g. if you have up to `carbon(7)`, but you delete `carbon(3)`, your next file will be saved as `carbon(3)`, which is mildly confusing).

## My changes

I added a few lines of code in the `save()` function within `Editor.js` that parse a timezone-adjusted ISO8601 time stamp (slightly formatted for readability). Of course, I also made it a toggle in the Settings so that it wasn't forced upon everyone.

## Result

Files are now saved in this format: `carbon yyyy-MM-dd hh-mm-ss.ext` (e.g. `carbon 2018-07-13 01-31-11.png`).

**P.S. I also added `.DS_Store` to the `.gitignore` since it was bugging me, but it's totally unrelated to the rest of this PR.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 cypress/videos
 cypress/screenshots
 .idea
+.DS_Store
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -142,9 +142,15 @@ class Editor extends React.Component {
 
   save({ id: format = 'png' }) {
     const link = document.createElement('a')
+    const timezoneOffset = (new Date()).getTimezoneOffset() * 60000
+    const timeString = (new Date(Date.now() - timezoneOffset))
+      .toISOString()
+      .slice(0, 19)
+      .replace(/:/g,'-')
+      .replace('T',' ')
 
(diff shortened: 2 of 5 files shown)


mfix22:
@erickzhao looks great! Can you actually do `carbon_yyyy-MM-dd_hh-mm-ss.ext` instead? (:

erickzhao:
@mfix22 Done! 🚀

PR #441, opened 2018-07-25
feat(save): add "copy" button

### LINKED ISSUE #49: Feature request: Copy Image to clipboard
It will be really great if there is copy to clipboard option.
Quite helpful when you have to share the image to slack - discord or any other chatting application which supports direct upload.

VinSpee:
add a button to the save menu called "copy", which opens the raw image data in the window so users
can directly copy it for easy pasting into a variety of apps

Closes #49

There is currently not browser support for copying actual images, so this seems to be a reasonable workaround. Any thoughts?

jul-25-2018 15-03-11

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -146,15 +146,9 @@ class Editor extends React.Component {
     const timestamp = this.state.timestamp ? `_${formatTimestamp()}` : ''
 
     return this.getCarbonImage({ format, type: 'blob' }).then(url => {
-      if (format === 'open ↗') {
-        link.href = url
-        link.target = '_blank'
-        document.body.appendChild(link)
-        link.click()
-        link.remove()
-        return url;
+      if (format !== 'open ↗') {
+        link.download = `carbon${timestamp}.${format}`
       }
-      link.download = `carbon${timestamp}.${format}`
       link.href = url
       document.body.appendChild(link)
       link.click()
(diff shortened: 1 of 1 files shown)


components/Editor.js
```diff
@@ -146,6 +146,10 @@ class Editor extends React.Component {
     const timestamp = this.state.timestamp ? `_${formatTimestamp()}` : ''
 
     return this.getCarbonImage({ format, type: 'blob' }).then(url => {
+      if (format === 'copy') {
+        window.location = url;
```
mfix22: Is there a way you can open it in a new tab? I think you can do this similar to how we handle the download case, but without setting "download".

components/Editor.js
```diff
@@ -36,7 +36,7 @@ const saveButtonOptions = {
   button: true,
   color: '#c198fb',
   selected: { id: 'SAVE_IMAGE', name: 'Save Image' },
-  list: ['png', 'svg'].map(id => ({ id, name: id.toUpperCase() }))
+  list: ['png', 'svg', 'copy'].map(id => ({ id, name: id.toUpperCase() }))
```
mfix22: can you call the button 'open ↗`

mfix22:
@VinSpee are you able to update this PR based on the comments above? 🙂

VinSpee:
Ooooh yep! I can today!

VinSpee:
👍🏻 check it out

mfix22:
With `target: "_blank"` ad-blockers will sometimes block URLs. We can either live with that, or just remove `target: "_blank"` (and probably the `↗` with it?). 

@briandennis @jakedex thoughts?

PR #458, opened 2018-08-08
Performance improvements

mfix22:
Time to first paint and interaction are both improved.
- Use `font-display: swap`
- Preload CSS links
- Compress all request payloads
- Inline Spinner CSS in critical path


components/Meta.js
```diff
@@ -4,8 +4,6 @@ import Reset from './style/Reset'
 import Font from './style/Font'
 import Typography from './style/Typography'
 
-import '../static/react-crop.css'
```
mfix22: I want to move all the custom themes to `/static` and just use `<links>`, removing `@zeit/next-css`

PR #466, opened 2018-08-11
encode text before reading the container width

### LINKED ISSUE #463: PNG cropped and SVG error when trying to saving/tweeting it (Golang code)
Hi folks,
I tried to create new image from my Golang code snippet.
When trying to export to PNG or SVG the image generated looks weird and seems an error happen.

### Exporting to png result
Image cropped
go-func *original image exported

### Exporting to svg result
image *i can't embed svg to github issue so this is the screenshot

After save the cropped image the editor also change my original code to, seems like, the escaped char version of my original code.

image

Code Snippet
```golang
package main

import "fmt"

func add(a, b int) (int, int, int) {
    return a, b, a + b
}

func main() {
    a, b, result := add(5,3)
    fmt.Printf("%d + %d equals to %d\n", a, b, result)
}
```

Hope this help the investigate issue.

mfix22:
This encodes the strings before measuring the width of the export container (which prevents the image from being cut off)

Closes #463

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -100,6 +100,14 @@ class Editor extends React.Component {
       return api.image(encodedState)
     }
 
+    if (format === 'png') {
+      document.querySelectorAll('.CodeMirror-line > span > span').forEach(n => {
+        if (n.innerText && n.innerText.match(/%\S\S/)) {
+          n.innerText = encodeURIComponent(n.innerText);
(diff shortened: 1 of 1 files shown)


mfix22:
@jakedex that issue is intentional. We need to escape those characters for dom-to-image to work on PNGs.

Issue #443, opened 2018-07-27
Prettier to format code on demand


farskid:
Thanks for the great tool. I user Carbon regularly and am a super fan of prettier as well. It would be nice to be able to integrate them together.

I've been tweaking around with prettier codebase lately and also got a demo running for a prettier rest api server and also a static version of prettier standalone UMD packages.

I've been so excited about this that I started to add it to carbon locally :)

This is just a basic demo from what I've got so far:

![carbon-prettier](https://user-images.githubusercontent.com/8332043/43343863-80bb693e-91f0-11e8-8096-775acc6e6127.gif)

Since prettier has a finite list of supported languages and technologies, I've been thinking about integrating dynamic parsers of prettier with the language that highlight.js detects from the editor.

Also prettier configurations might be good to be added to the settings dropdown menu.

I'd be happy to send a WIP PR if this looks good enough 👍

mfix22:
@farskid this looks cool, however I'm not sure we want to start implementing all of these editor features within Carbon. Yes Prettier is pretty easy to maintain, but we are not trying to make Carbon into another CodeSandbox (@compuives has you covered there 🙂) and could create a slippery slope.

farskid:
Totally fair.

My use case on a daily basis is such I use carbon a lot! and non pretty indented code pasted in carbon editor bugs me sometimes. I used to use my local projects or code sandbox for prettifying the code sample, but having in inside the carbon sounds like an enormous win 👍 

I'd be more than happy to help with the integration if you want this to happen sometimes in future :)

briandennis:
@farskid after talking some more with @mfix22, we think you make a good case for this. I'll defer to @jakedex as far as what the interface should look like, but go ahead and submit a WIP PR if you like 😃

farskid:
@briandennis Super Cool! I'm on it. will send a WIP very soon. Thanks for the feature people :)

PR #469, opened 2018-08-12
WIP: integration with prettier

farskid:
Following the discussion issue on https://github.com/dawnlabs/carbon/issues/443

## Description

For now it's only a naive simple formatCode using prettier's standalone package. It only uses Babylon parser.

Plan for changes:

* Smart parser switch: If the language is *auto* it should use babylon as default parser. Based on the supported list of languages built-in to the prettier and the detected language on Carbon, it should switch the parser if required.

* Showing an alert/toast in case format can't be done (prettier can be tricky sometimes)

* Adding prettier's options available in the settings dropdown

* Do the formatting in a web worker in case the code is a bit long.

@mfix22 @briandennis

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -217,13 +217,11 @@ class Editor extends React.Component {
   }
 
   format(code) {
-    formatCode(code)
-      .then(formatted => {
-        this.setState({ code: formatted })
-      })
-      .catch(err => {
-        console.error(err)
-      })
+    try {
+      this.setState({ code: formatCode(code) })
+    } catch (err) {
+      //
+    }
   }
 
   render() {
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -30,7 +30,7 @@ import {
   DEFAULT_SETTINGS
 } from '../lib/constants'
 import { serializeState } from '../lib/routing'
-import { getState } from '../lib/util'
+import { getState, formatCode } from '../lib/util'
 
 const saveButtonOptions = {
   button: true,
@@ -62,6 +62,7 @@ class Editor extends React.Component {
     this.resetDefaultSettings = this.resetDefaultSettings.bind(this)
     this.getCarbonImage = this.getCarbonImage.bind(this)
     this.onDrop = this.onDrop.bind(this)
+    this.format = this.format.bind(this)
 
     this.setOffline = () => this.setState({ online: false })
     this.setOnline = () => this.setState({ online: true })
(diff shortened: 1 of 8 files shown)


lib/util.js
```diff
@@ -40,3 +42,16 @@ export const fileToDataURL = blob =>
     reader.onload = e => res(e.target.result)
     reader.readAsDataURL(blob)
   })
+
+export const formatCode = code =>
+  new Promise((resolve, reject) => {
+    try {
+      const formatted = prettier.format(code, {
```
mfix22: If `prettier.format` is synchronous please make `formatCode` just return the code, not a promise.

components/Editor.js
```diff
@@ -251,15 +262,24 @@ class Editor extends React.Component {
...
-                  onClick={this.upload}
-                  title={this.state.uploading ? 'Loading...' : 'Tweet Image'}
-                  color="#57b5f9"
-                  style={{ marginRight: '8px' }}
-                />
-              )}
+              <Button
+                className="tweetButton"
+                onClick={() => this.format(this.state.code)}
+                title={this.state.formatLoading ? 'Formatting...' : 'P'}
```
mfix22: Remove `state.formatLoading` since formatting is synchronous.

lib/util.js
farskid: How should we handle format errors if it's sync? return Error object in the catch block and delegate error handling in the Editor?

mfix22:
@jakedex What do you think about making the `prettier` button be the actually [Prettier logo](https://prettier.io/icon.png), or a black and white version of it.

components/Editor.js
farskid: Done.

mfix22:
screenshot 2018-08-13 14 04 46

Obviously not final but just for inspo

farskid:
@mfix22 Should I work on the parser switch and re open this PR with that?

farskid:
@mfix22 Not long at all. Should go into the constants and switch pasrer based on lang. I can open new pr.

farskid:
It's more convenient to send it to another 3rd party API. I already created a prettier-server module which is doing this parser switching. I just don't know if you're interested in using it.

About the parser sizes, the babylon parser we're using right now as default is 185KB, graphql parser is 36KB, flow parser is 1.4MB, typescript parser is 2MB, ...
They're surprisingly huge!

mfix22:
As long as we code split the custom parsers out, I’m fine with loading it
in the browser.

Which I guess will make formatCode async again 😂.

farskid:
@mfix22 LOL, yeah I guess 😂. So I'll just work on it on a separate branch then and mention you. BTW, when is this merged branch gonna be deployed? with the next release?

farskid:
LOL, Yeah I guess it should be async again 😂. BTW, when is branch gonna be deployed? with the next release?

mfix22:
Yup! But I want to get a few details right before releasing. Like the code splitting to keep our bundle a little smaller.

PR #479, opened 2018-08-19
Make carbon a stateless component, move highlight.js calc into render

mfix22:


mfix22:
@briandennis is it worth also memoizing `handleLanguageChange`?

PR #486, opened 2018-08-22
restore offline

hanford:
@mfix22 restore carbon offline!

Sort of weird and something I'll need to document better, but it seems like `next-offline` is requiring that the `exportPathMap` is an async function. It makes sense in the next-offline context, because we're needing to do some weird filesystem stuff, but I don't want to diverge from the `next.js` spec, so when time permits, it's something I'll remove. And I can also ensure that `carbon` is updated as well.

Let me know if you'd like any other style/lint modifications, happy to update the PR.

(just to be sure I simply ran `yarn build && yarn export && cd out && now` and it appeared to be working. 

Here is a screenshot:
image

And a now environment:
https://out-bocovxgwiy.now.sh

Cheers!

https://github.com/hanford/next-offline/pull/19#issuecomment-414841537


mfix22:
@hanford awesome thanks! Just for future reference, the integration in the comment above always deploys a new version of Carbon, so you don't have to do so manually 👍

mfix22:
@hanford actually check out the new deployment here: https://carbon-gyiszlzous.now.sh/. The Service worker failed to register there too . . . can you take a quick look?

hanford:
@mfix22 very weird! I'll try to take a look today

mfix22:
@hanford could this be related to how the content is served in the OSS version of Now versus the paid? That is one difference between our builds . . .

hanford:
@mfix22 hmm, I don't know if there is a difference though the mime type bug is a little weird.

Since it's your now account, can you go to `https://carbon-gyiszlzous.now.sh/` and verify that a `service-worker.js` file doesn't exist?

mfix22:
@hanford confirmed. No `service-worker.js` file exists.

mfix22:
@hanford that worked. `next export` automatically copies over all the files in `/static`, so it might be worth putting the `service-worker.js` file in `/static` . . . what do you think?

mfix22:
@hanford https://github.com/hanford/next-offline/issues/47
if you want to go through with this, I can submit the PR

hanford:
@mfix22 PR would be awesome!

PR #488, opened 2018-08-26
Convert ids to classNames, use refs

mfix22:


payers1:
@mfix22 I think the api needs to be redeployed to `now`. Logs say the last deploy was August 23rd, and this change was August 27th I think. Puppeteer is still looking for `#export-container` and therefore returning the error `Error: Evaluation failed: TypeError: Cannot read property 'offsetWidth' of null`

mfix22:
Thanks @payers1 I will take a look immediately

PR #493, opened 2018-08-31
tune component updates and bump codemirror polling to hit 60 Hz

mfix22:
This change is not worth doing unless you can sense a noticeable improvement in update perf when typing. 

### Before
carbon-before

### After
carbon-after

10x faster React updates ^


mfix22:
@jakedex @briandennis

Issue #450, opened 2018-08-04
Create multiple beautiful default themes to choose from


mittalyashu:
### Expected Behavior
While playing around with the settings, it is possible that we might end up creating something like this, which looks horrible.

![carbon](https://user-images.githubusercontent.com/29014463/43673067-f4a232ee-97d9-11e8-8a30-c2f022fd0ee8.png)

That's why we need help.

### Actual Behavior

Generate a beautiful random settings.

mfix22:
We do want to support more beautiful default options, but definitely not random 😝 
Maybe close this issue, and create a new one? Or just update the titles and such.

mittalyashu:
I hope you have understood what I mean to say.

mittalyashu:
https://www.useloom.com/share/9a9d27aceb584264bd9b9866f74aca76

mfix22:
@mittalyashu totally understood! We do want to create a bunch of beautiful default themes that set the font, background, and color-theme, padding etc so users can quickly choose one that works best for them. But we don't want it to be "random" per se (:

mittalyashu:
> We do want to create a bunch of beautiful default themes, but we don't want it to be "random".

Got it.

Jaredk3nt:
I'm interest in taking on this issue,
Just to make it clear is the scope adding a 'templates' drop down menu as well as creating default themes to add to the menu?
Also I'm assuming there will some approval process on the default themes?
Thanks!

mittalyashu:
> Just to make it clear is the scope adding a 'templates' drop down menu as well as creating default themes to add to the menu?

Yes, a predefined theme should include all the settings which carbon currently provide and we should be able to override the theme with custom setting if we want.

Jaredk3nt:
Awesome, I would love to be assigned to this!

mfix22:
@Jaredk3nt feel free to open a WIP PR and I can give you feedback as you go. I figured posting screenshots of the theme would be a good way to verify which themes we like before finalizing on the selection UI.

@jakedex and I can help out the theme menu

Jaredk3nt:
Making a bunch of examples here are a couple that I think are good starts. Feedback is appreciated! I looked through a bunch that people have posted places to get a feel for the kinds of themes people are already using and are doing well.

![carbon](https://user-images.githubusercontent.com/6362515/45859021-ca48ac00-bd25-11e8-8839-305d115bac56.png)
[...]

PR #507, opened 2018-09-22
Adding default themes (#450)

Jaredk3nt:
image

## Description
Added a dropdown where the user can select templates that will change multiple settings. Added starter default templates in constants to select from.


Jaredk3nt:
[WIP] would love some feedback on the placement of the dropdown, is it ok if it spills over? Also need to add more themes once I get feedback on those (they are attached to the issue)

mittalyashu:
@mfix22 Can we get access to the deployment so that we can test and give feedback.

mfix22:
Deployment: https://carbon-lzcqluhqdf.now.sh/

(all of the PR deployments are available below the Travis CI job)

lib/constants.js
```diff
@@ -521,3 +521,32 @@ export const DEFAULT_SETTINGS = {
   squaredImage: false,
   timestamp: false
 }
+
+export const TEMPLATES = [
+  {
+    id: 'carbon-default',
+    name: 'Carbon Default',
```
mfix22: `carbon-default` -> `default`
`'Carbon Default' -> `'Default'`

PR #519, opened 2018-10-03
Add N-Triples language support

roddolf:
## Description
- Add N-Triples language support

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -318,11 +318,6 @@ export const LANGUAGES = [
     short: 'mysql'
   },
   {
-    name: 'N-Quads',
-    mode: 'n-quads',
-    mime: 'application/n-quads'
-  },
-  {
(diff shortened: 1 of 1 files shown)


lib/constants.js
```diff
@@ -317,6 +317,16 @@ export const LANGUAGES = [
     mime: 'text/x-mysql',
     short: 'mysql'
   },
+  {
+    name: 'N-Quads',
+    mode: 'n-quads',
+    mime: 'application/n-quads'
+  },
+  {
```
mfix22: Can we only add one of these as to try not and overload the language selector? 

I would prefer N-Triples, since according to the CodeMirror docs:
> The N-Triples mode also works well with on N-Quad documents.

roddolf: Sure, I understand your point.
I have made the changes.

Issue #439, opened 2018-07-21
Emojis and Chinese characters are not rendered in the output


dmcrodrigues:
## Expected Behavior

<img width="338" alt="screen shot 2018-07-21 at 16 51 42" src="https://user-images.githubusercontent.com/1098602/43037569-6e0a2b84-8d06-11e8-899f-d11d5d3ba78a.png">

## Actual Behavior

![carbon-12](https://user-images.githubusercontent.com/1098602/43037574-74ed2532-8d06-11e8-9f78-408b59738a69.png)

mfix22:
Hey thanks for filing this. What browser were you using?

dmcrodrigues:
Hey @mfix22, I'm using Safari 11.1.2 (13605.3.8).

mittalyashu:
It works for me, I am using Chrome: 68.0.3440.84

![carbon 1](https://user-images.githubusercontent.com/29014463/43673124-b4cfea3e-97da-11e8-87dc-3462f7a11711.png)

mfix22:
@mittalyashu the failure case is for Safari

mfix22:
I believe this is fixed in `puppeteer@1.6.*`, however, we need to wait for #451 to be fixed before we can upgrade our puppeteer version.

ajarouih:
It works for me with the same configs as yours try this emoji as copy/paste  👌

mfix22:
@ajarouih it works in all browsers except Safari

sdabhi23:
@mfix22 as #451 has been fixed, can I help fix this one?

mfix22:
@sdabhi23 Absolutely! we would appreciate your help on this one

mfix22:
Just pushed an update the adds the correct fonts for emojis and simplified Chinese!

PR #526, opened 2018-10-06
Update puppeteer to latest version

sdabhi23:
Update version of puppeteer to address issues with emojis in the output.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -17,7 +17,7 @@
     "morgan": "^1.8.2",
     "morphmorph": "^0.1.2",
     "now-logs": "^0.0.7",
-    "puppeteer": "1.7.0",
+    "puppeteer": "1.9.0",
     "twit": "^2.2.9",
     "unsplash-js": "^4.8.0"
   }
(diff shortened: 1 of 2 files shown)


sdabhi23:
@mfix22 I do not have an Apple device to test this. Can you please check and confirm if the issue has been fixed?

mfix22:
@sdabhi23 absolutely! Is this to fix issue #439?

sdabhi23:
@mfix22 yes it was to fix #439, and sorry for not testing it out properly.

mfix22:
I upgraded it in the other PR, and emojis are still not working for me 🤷‍♂️ I am still at a loss on this bug (:

sdabhi23:
@mfix22 I am not able to exactly understand the flow of the app, but from whatever I have understood, I guess it's not a problem with the API, but rather the problem is the way in which Safari encodes the emoji characters.

What I have understood till now is that the conversion from code to image is happening on the server, which should be independent of the client from which the site is being accessed.

PR #529, opened 2018-10-07
update contributor tables

racaljk:


racaljk:
a tool for contributor table generating may effective since we don't need to edit every readme doc

mfix22:
@racaljk sorry yeah you don't need to manually enter it. you can just run `yarn contrib:add` and then your user name /: Sorry I didn't tell you that sooner.

PR #551, opened 2018-10-16
Fix broken link on paragraph #016

kant:
## Description

* Update link on #016
* Plus minor proposals


mfix22:
@kant thank you 🙏

PR #571, opened 2018-11-02
Code cleanup.

FozzieHi:
Just simplified a few statements and changed inline variables.

### DIFF
--- a/lib/custom/modes/apache.js
+++ b/lib/custom/modes/apache.js
@@ -8,8 +8,8 @@ const CodeMirror = require('codemirror')
 CodeMirror.defineMode('apache', function(config) {
   return {
     token: function(stream, state) {
-      var sol = stream.sol() || state.afterSection
-      var eol = stream.eol()
+      const sol = stream.sol() || state.afterSection
+      const eol = stream.eol()
 
       state.afterSection = false
 
--- a/lib/custom/modes/nimrod.js
+++ b/lib/custom/modes/nimrod.js
@@ -36,16 +36,16 @@ CodeMirror.defineMode('nimrod', function(conf, parserConf) {
     '.\\.',
     '.'
   ]
-  var operators = new RegExp(
+  const operators = new RegExp(
     ops
       .map(function(op) {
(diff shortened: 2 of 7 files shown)


lib/custom/modes/nimrod.js
```diff
@@ -36,16 +36,16 @@ CodeMirror.defineMode('nimrod', function(conf, parserConf) {
     '.\\.',
     '.'
   ]
-  var operators = new RegExp(
+  const operators = new RegExp(
```
mfix22: Please remove `/*eslint-disable*/` from this file and fix the errors

next.config.js
```diff
@@ -2,7 +2,7 @@
 // const withOffline = require('next-offline')
 
 module.exports = (/* phase  { defaultConfig } */) => {
-  const config = {
+  return {
```
mfix22: Please leave this as is

lib/custom/modes/apache.js
```diff
@@ -8,8 +8,8 @@ const CodeMirror = require('codemirror')
 CodeMirror.defineMode('apache', function(config) {
   return {
     token: function(stream, state) {
-      var sol = stream.sol() || state.afterSection
-      var eol = stream.eol()
+      const sol = stream.sol() || state.afterSection
```
mfix22: Please remove `/* eslint-disable */` from this file and fix errors before merging

FozzieHi:
The eslint disables were already present, I did not add them. Some of the eslint errors would require removing function parameters, do you still want me to do this? @mfix22

mfix22:
I know they were there 🙂Those files were just copy pasted from somewhere, so if we are going to make edits to them i want to treat them as actually part of the code base, so they should be linted appropriately.

You are going to need to update the linting script to change `lib/*.js` to `lib/**/*.js` too. I appreciate it 👍

mfix22:
@FozzieHi are you able to rebase this on master and make the requested changes? Otherwise I will close this PR since it is just for clean up (which we appreciate) 🙂

PR #585, opened 2018-11-20
Dropdown updates

raboid:
image

image

image

## Description
- Add icons to Theme and Language dropdowns
- Remove BG label from BackgroundSelect
- Make border-widths 2px when BG & Settings are open


mfix22:
@raboid should the DropdownIcon's open and close the dropdown? I'm fine either way, but if they aren't supposed to, then let's remove the `cursor: pointer` from them

PR #595, opened 2018-12-05
Add preset feature

### LINKED ISSUE #450: Create multiple beautiful default themes to choose from

raboid:
## Description
- Adds the functionality to select a preset and have those settings be applied

Closes #450


components/Settings.js
```diff
@@ -360,6 +381,12 @@ class Settings extends React.PureComponent {
         </div>
         <div className="settings-settings">
           <WindowPointer fromLeft="15px" />
+          <Presets
+            show={showPresets}
+            toggle={this.togglePresets}
+            presets={PRESETS}
```
mfix22: @raboid let's add `PRESETS` to `state` in `Editor` and then pass that value in as a prop into `Settings`. That way we can support adding/removing presets easily.

components/Editor.js
```diff
@@ -329,6 +335,8 @@ class Editor extends React.Component {
               {...config}
               onChange={this.updateSetting}
               resetDefaultSettings={this.resetDefaultSettings}
+              updatePreset={this.updatePreset}
+              presets={presets}
```
mfix22: This line is not needed since it is included in `config` above

components/Settings.js
```diff
@@ -202,94 +202,114 @@ const MenuButton = React.memo(({ name, select, selected }) => {
...
-            </div>
+          {presets.map(preset => (
+            <button
+              key={preset.name}
+              className="settings-presets-preset"
+              style={{ backgroundColor: preset.backgroundColor }}
+              onClick={() => update(preset)}
+            >
+              {preset.custom ? (
+                <button className="settings-presets-remove" onClick={() => remove(preset.name)}>
```
mfix22: Can you hook up `remove` to remove that preset from the Editor's state? (:

raboid: Yeah I was going to wait until we had custom presets to remove, but I'll do it now

components/Settings.js
```diff
@@ -203,33 +204,109 @@ const MenuButton = React.memo(({ name, select, selected }) => {
...
+        )}
         <button className="settings-presets-arrow" onClick={toggle}>
           {show ? <Arrows.Up /> : <Arrows.Down />}
         </button>
       </div>
       {show ? (
         <div className="settings-presets-content">
-          {presets.map((preset, i) => (
-            <button
+          {presets.slice(0, customPresetsLength).map((preset, i) => (
```
raboid: I felt this way was more understandable than doing a single loop and figuring out where to insert the divider. I don't like how dependent this code is on array indexes, but I'm not sure using id's and then having to sort them is any better.

components/Settings.js
```diff
@@ -391,33 +464,74 @@ class Settings extends React.PureComponent {
...
+        this.props.selectPreset(0)
+        this.savePresets()
+      }
+    )
+  }
+
+  savePresets = () =>
+    savePresets(
+      localStorage,
+      this.state.presets.slice(0, this.state.presets.length - DEFAULT_PRESETS.length)
```
raboid: Another spot where using index's comes up. This could be presets.filter(preset => preset.custom) which is probably more understandable?

mfix22: Yes please make this change 👍

components/Settings.js
mfix22: Again let's use `filter(p => p.custom)` and `filter(p => !p.custom)` here instead of `slice` 👍

components/Settings.js
```diff
@@ -340,8 +462,76 @@ class Settings extends React.PureComponent {
...
+      }
+    )
+  }
+
+  createPreset = async () => {
+    const preset = this.getSettings()
+
+    preset.custom = true
+
+    preset.icon = await this.props.getCarbonImage({ format: 'png', squared: true })
```
mfix22: Can you also try passing an `exportSize` < 1 to see if we can't cut down on localStorage size? 🙂

lib/util.js
```diff
@@ -36,12 +38,22 @@ export const unescapeHtml = s => {
 
 export const parseRGBA = obj => `rgba(${obj.r},${obj.g},${obj.b},${obj.a})`
 
-export const getState = morph.compose(
+export const getSavedSettings = morph.compose(
```
mfix22: Can we change this to `getSettings` (and `getPresets` accordingly)? The reason being I like the parallelism between `getX` and `saveX`

jakedex:
Ahh this is slick! Create preset works like a charm 💫

A few nits from me:

A background change should register as a "change" to the preset (in the same way that changing padding, etc does)

.settings-presets-applied
- display: flex
- justify-content: space-between

"create +"
- cursor: not-allowed when disabled
- line-height: 1

.settings-presets-arrow
- display: flex

jakedex:
Just noticed an issue... the preset remains selected when "Reset Settings" is clicked. I think we should do the following: 

- [ ] Create preset from current default theme (reset settings state)
- [ ] This "default" preset should be selected when a user resets their settings or visits carbon for the first time

jakedex:
Also noticed an issue with the "preset applied" banner stacking, in that the banner remains after clicking undo once and clicking it again does nothing but close the banner. I think we want to eliminate the stacking altogether — i.e. clicking undo brings you back to your previous state and hides the banner.

jakedex:
We should maintain scroll position when selecting presets... 

https://cl.ly/9d7cd48b4969

components/Settings.js
```diff
@@ -340,8 +456,64 @@ class Settings extends React.PureComponent {
     }
   }
 
+  handleChange = (key, value) => {
+    this.props.onChange(key, value)
+    this.setState({ previousSettings: null })
+  }
+
+  getSettingsFromProps = () => omit(omitBy(this.props, v => typeof v === 'function'), ['preset'])
```
mfix22: Can you just make this a single `omitBy`? 🙂 Sorry to be a pain

lib/util.js
```diff
@@ -65,3 +78,5 @@ export const formatCode = async code => {
     singleQuote: true
   })
 }
+
+export const omit = (object, keys) => omitBy(object, (_, k) => keys.includes(k))
```
mfix22: Can you use `.indexOf(k) > -1`? `Array.includes` is not supported in all browsers yet (https://caniuse.com/#feat=array-includes)

raboid: Ah yeah no problemo, I'm so used to just babel/polyfilling my way out of those concerns lol

mfix22: Hahaha yeah i can tell with all your class instance functions. I dig it, but... 🙂

raboid:
07b932f adds a presetApplied state variable per @jakedex request that the preset that is first selected doesn't show the 'Preset Applied' bar. Also when you create a preset it should be selected, but also not show the 'Preset Applied' bar

components/Settings.js
```diff
@@ -286,16 +262,171 @@ const MenuButton = React.memo(({ name, select, selected }) => {
...
         }
       `}
-      </style>
-    </div>
+    </style>
+  </div>
+))
+
+const Presets = React.memo(
+  React.forwardRef(
```
mfix22: @raboid you don't need to use `forwardRef` unless the wants to pass the ref in the `ref` prop (https://reactjs.org/docs/forwarding-refs.html) You can just remove it in these case since you are using the `contentRef` prop instead

raboid: Oh interesting I had thought it was needed in general to pass a ref created in the parent to a child. Thanks!

PR #604, opened 2018-12-14
Reduce updates caused by typing

### LINKED ISSUE #603: Improve rerender performance when typing
- [ ] Remove https://github.com/dawnlabs/carbon/blob/master/components/Editor.js#L116 and instead call `this.props.onUpdate` on state updates (you can create a new function called `this.updateState` that wraps `setState`)
- [ ] Prevent things from re-rendering when the URL changes (i.e. the React state should be the source of truth, not the URL)
  - **Idea**: add `componentShouldUpdate = () => false` to https://github.com/dawnlabs/carbon/blob/master/pages/index.js#L12
- [ ] Use the React Profiler (https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) and/or Chrome profiler to investigate and then remove unnecessary re-renders when typing in the editor textarea

raboid:
Resolves #603 

## Description
- Prevents index page from updating
- Implements `shouldComponetUpdate` for `ExportMenu` to prevent unneeded updates caused by the `router` prop changing.
- Replaces `componentDidUpdate` with an `updateState` function in `Editor`

### DIFF
--- a/components/ExportMenu.js
+++ b/components/ExportMenu.js
@@ -31,6 +31,7 @@ const CopyEmbed = withRouter(
             .copy-button {
               display: flex;
               flex: 1;
+              flex-basis: 68px;
               justify-content: center;
               align-items: center;
               padding: 12px 16px;
--- a/components/ExportMenu.js
+++ b/components/ExportMenu.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import shallowCompare from 'react-addons-shallow-compare'
 import enhanceWithClickOutside from 'react-click-outside'
 import { withRouter } from 'next/router'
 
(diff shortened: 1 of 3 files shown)


components/ExportMenu.js
```diff
@@ -17,11 +18,18 @@ const toIFrame = url =>
 </iframe>
 `
 
-class ExportMenu extends React.PureComponent {
+class ExportMenu extends React.Component {
   state = {
     isVisible: false
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
```
mfix22: Sorry to get back to you late on this @raboid can you avoid using `react-addons-shallow-compare`? I want to actually eliminate that dependency from this code base.

An easy alternative would be to extract the pieces that are `this.props.router.asPath` and then use [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) to only update that piece when `this.props.router.asPath` updates

PR #605, opened 2018-12-15
Add TSX language

jrobusto:
Adds the `TSX` language format for Typescript-based React projects. 
Fits in well with existing support for Typescript and JSX languages.

Snippet Example

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -431,6 +431,12 @@ export const LANGUAGES = [
     short: 'typescript'
   },
   {
+    name: 'TSX',
+    mode: 'jsx',
+    mime: 'text/typescript-jsx',
+    short: 'tsx'
+  },
+  {
     name: 'VB.NET',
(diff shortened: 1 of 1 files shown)


lib/constants.js
```diff
@@ -430,6 +430,12 @@ export const LANGUAGES = [
     mime: 'application/typescript',
     short: 'typescript'
   },
+  {
+    name: 'TSX',
+    mode: 'jsx',
+    mime: 'text/typescript-jsx',
```
mfix22: I'm fairly certain you do not need `mime` field here. Please remove 👍

jrobusto: I'm not sure on the verbage exactly, but it seems to distinguish `text/jsx` and `text/typescript-jsx` MIME types within the single `JSX` mode. 

https://codemirror.net/mode/jsx/index.html

PR #607, opened 2018-12-15
Caution about the use of images in a post

rachelcarmena:
## Description
* Adding more information in About page about the use of images with source code. This warning is motivated after talking to Juanjo Montiel, a blind developer, who finds it impossible to read the source code if it's within images. Furthermore, there are other visual disabilities for which text within images cannot be enlarged correctly. On the other hand, the source code cannot be copied. I find it interesting to raise awareness of it.
* Updating the section `Export/Sharing` in README.md with the same information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ Once you've got all of your code into Carbon, you can customize your image by ch
 
 After you've customized your image you can either Tweet a link to the image, or save it directly.
 
+If you use the &apos;Tweet&apos; button, Carbon will automatically make your image accessible. However, if you want to manually tweet your carbon image, please check out [how to make your Twitter images accessible](https://help.twitter.com/en/using-twitter/picture-descriptions).
+
+If you include a carbon image in a post, the source code will be invisible for assistive technology, it will not be possible to enlarge it or copy it, etc. Please, think about adding another element with the source code as text. For example, you can add [a HTML Details Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) below the image.
+
 ## Community
 Check out these projects our awesome community has created:
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ After you've customized your image you can either Tweet a link to the image, or 
 
 If you use the &apos;Tweet&apos; button, Carbon will automatically make your image accessible. However, if you want to manually tweet your carbon image, please check out [how to make your Twitter images accessible](https://help.twitter.com/en/using-twitter/picture-descriptions).
 
-If you include a carbon image in a post, the source code will be invisible for assistive technology, it will not be possible to enlarge it or copy it, etc. Please, think about adding another element with the source code as text. For example, you can add [a HTML Details Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) below the image.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,10 @@ Once you've got all of your code into Carbon, you can customize your image by ch
 
 After you've customized your image you can either Tweet a link to the image, or save it directly.
 
+If you use the &apos;Tweet&apos; button, Carbon will automatically make your image accessible. However, if you want to manually tweet your carbon image, please check out [how to make your Twitter images accessible](https://help.twitter.com/en/using-twitter/picture-descriptions).
+
+If you include a carbon image in a post, the source code will be invisible for assistive technology, it will not be possible to enlarge it or copy it, etc. Please, think about adding another element with the source code as text. For example, you can add [a HTML Details Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) below the image.
```
mfix22: I believe for images an `alt` attribute is sufficient for making images screen-reader accessible, no? 🙂 I would love to encourage people to do that. 

Alternatively, users of Carbon can add `title` attributes to their images to also add screen-readability. Do you think this is an equivalent alternative?

rachelcarmena: Hi @mfix22, I understand you, because I thought the same at the beginning, but we must open our mind to other types of disabilities and other types of assistive technology. 

That solution would be valid for screen readers. Furthermore, ARIA provides `aria-describedby` and `aria-details` for facilitating how to add a description to an image.

However, in this case, it's not a simple description. It's the text within the image. There are other visual disabilities for which text within images cannot be enlarged correctly, the font cannot be changed, the contrast cannot be changed, etc.

That's the reason why it's important to offer another element with the source code as text. 

It might appear difficult to take into account all the disabilities and all the types of assistive technology, although I think it's easy in this case. Text within an image is not accessible, it's only necessary to offer the source code as text in an additional element.

You have made a great work, because source code is beautiful with Carbon. I thought about a HTML Details Element, because it doesn't interfere too much with the visual aspect (its content is only visible when it's toggled into an "open" state). Maybe there are more options, but my goal is to raise awareness of it.

Thanks for your time, @mfix22 !

mfix22:
Thanks @rachelcarmena! 🙂

rachelcarmena:
Thanks to you, @mfix22, thanks for your understanding. 

I've made a lot of mistakes, because of a lack of knowledge. I've introduced barriers in a technological world. Those barriers that can be found in the real world. However, we can create a technological world without barriers. 

Tim Berners-Lee said:
> The power of the Web is in its universality. 
> Access by everyone regardless of disability is an essential aspect.

We have a great responsibility as developers in order to give that power to the Web. And I think we can get it :muscle:

Thanks @mfix22 !!!

mfix22:
@rachelcarmena absolutely! I was only asking for my own curiosity's sake (: Would love to help this cause as best I can 🙂

rachelcarmena:
Oh, I'm sure about it, @mfix22. Sorry for my "speech" :sweat_smile: 

Don't stop asking questions. 
I asked a lot of questions until I had this proposal for you.

Thanks again for your time!! :hugs:

mfix22:
Thanks for your contributions to Carbon 👍 we appreciate it

Issue #232, opened 2018-02-13
Upload your own code font


jameslittle230:
Could be cool to upload your own monospace font to be used in the output, for those who have monospace fonts that aren't available for free.

GavinRay97:
@mfix22 
Something like this maybe? :smile: 

Video:
https://i.imgur.com/Uw3Jsj4.gifv

![](http://i.imgur.com/kpfO8P3.png)

mfix22:
@GavinRay97 that looks great too! Can you push that up as a second PR to your custom theme's one? Also can you open up a PR for that too so that the community can see both?

GavinRay97:
@mfix22 Awesome. I uploaded both, but I'm not sure what branch to base the PR off of.

https://github.com/GavinRay97/carbon/tree/custom-fonts
https://github.com/GavinRay97/carbon/tree/custom-themes

mfix22:
Just released the ability to upload your own font! Will add the ability to save them to reuse later.

PR #613, opened 2019-01-03
add font operator mono

Seaony:
Add Operator mono font


mfix22:
This is a licensed font, and I'm fairly certain that using it from this CDN would be a breach of the license. Therefore we cannot support adding this font. 

However, we soon want to support adding your own fonts, in which case you can upload your own copy of Operator (#232)

PR #622, opened 2019-01-08
Cleanup

raboid:
## Description
- Make the code more DRY by creating Popout, Input, Button, ColorPicker components.
- Cleanup CSS and some class names
- Use util/toggle more

These changes made the upcoming Custom Theme feature much nicer to implement. I also think we're getting to a spot where it'd be great to add styled-components.

Also, nice to see there is actually more deletions in this PR then additions lol.


components/Button.js
```diff
@@ -1,39 +1,62 @@
...
+  border,
+  center,
+  large,
+  style = {},
+  flex = 1,
+  padding = 0,
+  margin = 0
+}) => (
+  <button onClick={onClick} className={className} disabled={disabled || notAllowed} style={style}>
+    {title && <span className="button-title">{title}</span>}
```
mfix22: What is the difference between title and children? It seems like this shouldn't be the job of the button component, but rather it should just take the children passed in and render them directly. Do you think having both is a better API?

components/Settings.js
```diff
@@ -1,15 +1,14 @@
...
 
 import ThemeSelect from './ThemeSelect'
 import FontSelect from './FontSelect'
 import Slider from './Slider'
 import Toggle from './Toggle'
-import WindowPointer from './WindowPointer'
+import Popout from './Popout'
+import Button from './Button'
 import { COLORS, DEFAULT_PRESETS } from '../lib/constants'
-import { getPresets, savePresets } from '../lib/util'
```
mfix22: Great catch! Is there an eslint rule to prevent duplicate imports like this?

components/Settings.js
```diff
@@ -171,109 +160,102 @@ const MiscSettings = React.memo(({ format, reset }) => {
...
+const removeButtonStyles = {
+  position: 'absolute',
+  top: '6px',
+  right: '6px',
+  width: '11px',
+  height: '11px',
+  borderRadius: '999px'
+}
+
+const Preset = React.memo(({ remove, apply, selected, preset }) => {
```
mfix22: While you are at it, can you move the Preset menu into its own file? 🙂

raboid: Yes agreed I was wanting to do that. I did that for the new Theme feature and it made everything much better. I planned to wait until after the Theme work is done though. This PR is mostly things I pulled out of that work.

components/Button.js
raboid: Yeah not needed. I kind of like not having to use a closing tag if it's just text as children, but that's the only benefit it's providing right now.

components/Settings.js
raboid: https://eslint.org/docs/rules/no-duplicate-imports

We could add that to our eslint, surprised it's not in the recommended rules.

mfix22: Please add it 👍

components/Button.js
mfix22: I would rather simplify the Button API as much as possible

raboid:
That incorrect behavior should be fixed now. I moved the popout logic into a HOC, which we can swap for a hook based solution when hooks become stable.

mfix22:
@raboid Why are we writing our own HOC, when `enhanceWithClickOutside` already gave use an HOC that worked correctly? I feel like this new abstraction adds unnecessary confusion/complexity. Is it just so we only need to have the `isVisible` state code in a single place?

PR #632, opened 2019-01-11
Custom themes

raboid:


components/Themes.js
```diff
@@ -0,0 +1,361 @@
...
+      this.props.updateTheme(DEFAULT_THEME.id)
+    } else {
+      this.setState({ themes: newThemes })
+    }
+  }
+
+  createTheme = () => {
+    const { themes, name, highlights } = this.state
+
+    const id = generateId()
```
mfix22: The `name` should be the `id` in this case

raboid: I didn't use the name to avoid the edge case where the name already exists.

components/Themes.js
```diff
@@ -0,0 +1,361 @@
...
+        <Dropdown
+          icon={themeIcon}
+          disableInput={isVisible}
+          inputValue={dropdownValue}
+          selected={dropdownValue}
+          list={dropdownList}
+          itemWrapper={ThemeItem}
+          itemWrapperProps={{ onClick: this.removeTheme }}
+          onChange={this.handleDropdown}
+          onOpen={isVisible && toggleVisibility}
```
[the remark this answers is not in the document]
raboid: So I need a callback to know when the dropdown is opened so that I can close this popout. Is there a better way for `Themes` to know when the dropdown is clicked open?

components/Themes.js
mfix22: Yeah I don't think we should allow creating custom themes with the same name 🙂

components/Themes.js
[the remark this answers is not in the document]
mfix22: Sounds good 👍

lib/constants.js
```diff
@@ -15,118 +15,503 @@ export const FONTS = [
...
+      keyword: '#f0dfaf',
+      meta: '#f0dfaf',
+      number: '#dcdccc',
+      operator: '#f0efd0',
+      string: '#cc9393',
+      variable: '#dcdccc'
+    }
   }
-]
+].map(t => ({ ...t, highlights: { ...DEFAULT_HIGHLIGHTS, ...(t.highlights || {}) } }))
```
mfix22: Do you still need this line?

mfix22:
@raboid can you disable the save button if there is no custom name? The default "Custom Theme 1" is great! But if people delete it it ends up looking funny:

screenshot 2019-01-18 14 12 38

lib/constants.js
```diff
@@ -15,118 +15,503 @@ export const FONTS = [
...
+  string: '#a11',
+  tag: '#170',
+  attribute: '#00c',
+  comment: '#a50'
+  //meta: '#555'
+}
+
+export const HIGHLIGHT_KEYS = Object.keys(DEFAULT_HIGHLIGHTS)
+
+// TODO: Add more theme highlights
```
mfix22: remove

lib/constants.js
raboid: No I can just go through and fill in the defaults in the themes that need them.

components/Themes.js
```diff
@@ -0,0 +1,391 @@
...
+            updateHighlight={this.updateHighlight}
+          />
+        )}
+        <style jsx>
+          {`
+            .themes {
+              position: relative;
+            }
+
+            :global(.CodeMirror) {
```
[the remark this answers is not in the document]
mfix22: Perfect 👍

jakedex:
This is looking really nice! A few things: 

- the color-picker dropdown width is wonky for me now? https://cl.ly/70f0ec0dc593
- custom theme isn't previewed in the editor while it's being edited
- `.color-picker-header` and `.settings-presets-header` should be `font-size: 12px`

PR #642, opened 2019-01-22
Use dynamic import components

mfix22:
@raboid do you think this is a good idea? Saves us quite a bit on initial bundle size. Can you investigate more low hanging fruit? For example I think we are shipping _all_ of lodash when we only need two functions


lib/constants.js
```diff
@@ -908,18 +909,11 @@ const unfold = (f, seed) => {
...
-        alreadyLoaded.add(language.mode)
-        language.custom
-          ? require(`./custom/modes/${language.mode}`)
-          : require(`codemirror/mode/${language.mode}/${language.mode}`)
-      }
-    }
-  )
+  const alreadyLoaded = new Set(['javascript'])
+
+  dynamic(() => import('./modes')(alreadyLoaded))
```
mfix22: I don't think you need to pass in `alreadyLoaded` you can just do:
```js
import('./modes').then(modes => {
  modes.forEach(l => alreadyLoaded.add(l.mode))
})
```
or something similar

mfix22: or rather wrap the whole import call in a check to see it it had been imported before.

PR #658, opened 2019-02-07
update meta

raboid:
## Description
- `Page` now only includes `MetaTags` and `Styles`


mfix22:
@raboid did you verify in the bundle that this is getting code-split appropriately for the `/about` page?

PR #683, opened 2019-02-23
Exported SVG is too large.

Joel-hanson:
The size of the SVG got large due to unwanted font-face data included in the SVG file. So we need only the Fonts-face data that is used for the currently export SVG file.

After removing the unwanted fonts the size SVG got decreased.
screenshot 2019-02-23 at 2 31 46 pm
screenshot 2019-02-23 at 2 31 57 pm

The files having the same code text with 2x size are in the following zip file.

[Archive.zip](https://github.com/dawnlabs/carbon/files/2896659/Archive.zip)

## Description
`components/Editor.js`
- Initialized fontFamily to store the current font-family chosen by the user
- Used replace with regex to remove the unwanted font-face data from the file.
`.all-contributorsrc`
- Added my GitHub details

## HOPE THIS DESCRIPTION HELPS ##


components/Editor.js
```diff
@@ -172,6 +174,14 @@ class Editor extends React.Component {
                   .replace(/%23/g, '#')
                   .replace(/%0A/g, '\n')
               )
+              // remove other fonts which are not used
+              .then(
+                dataUrl =>
+                  (dataUrl = dataUrl.replace(
```
mfix22: Is it possible to use the `dom-to-image`'s `filter` option (https://github.com/dawnlabs/carbon/pull/683/files#diff-24387bc9bda6a21543364a64ae3b83c1R152) in order to remove unwanted fonts? If you can even separate them into separate `style` nodes if that make's it easier

Joel-hanson: Thank you @mfix22 for your comments and suggestions. It's a cool suggestion, and I have never thought of this.

The issues that I faced on implementing the suggestion -

1. I tried to fix it with the `filter`. But the styles nodes (having font-face) are appended by the `embedFonts` in `dom-to-image` (`embedFonts` having the `newFontFaces` in `dom-to-image` reads all the `document.styleSheets` to get the font-faces.), Thus the `filter` did not come across the style nodes, so was not able to use `filter` to prevent unwanted style nodes.
2. And there is no option to update the embedded fonts in `dom-to-image`.

This issue to **Optimize downloading webfonts** are being discussed [https://github.com/tsayen/dom-to-image/issues/37](here) in the [https://github.com/tsayen/dom-to-image](dom-to-image)

We can make this suggestion happen by updating the https://github.com/tsayen/dom-to-image/blob/master/src/dom-to-image.js to get only the fonts needed.

Joel-hanson: Please correct me if any thing that i mentioned above is wrong.

components/Editor.js
```diff
@@ -172,6 +174,14 @@ class Editor extends React.Component {
...
+                    new RegExp('@font-face { font-family: (?!"*' + fontFamily + ').*?}s?', 'g'),
```
mfix22: What is the `s?` part of the regex for?

components/Editor.js
mfix22: 👍

components/Editor.js
Joel-hanson: \s matches a space, a tab, a line break, or a form feed.

Joel-hanson: I think we won't be needing that.

Joel-hanson:
I have updated the regex having `s?`, by removing it. Have tested the SVG download and works fine

components/Editor.js
mfix22: It's just in cases in the future the output is different, in which case `\s` is safer. To be fair, I don't think it will ever change, especially b/c `dom-to-image` hasn't gotten released in months.

mfix22: Also, don't be sorry 🙂 we are working on this together!

mfix22:
Thanks @Joel-hanson 🙏

Joel-hanson:
Thanks a lot @mfix22 for the last changes and for merging my pull request.

mfix22:
I really appreciate people going and adding optimizations to Carbon like this! If you have other ideas, feel free to post them in the issues!

PR #692, opened 2019-03-05
Read config from gist (optional)

mfix22:


components/Editor.js
```diff
@@ -293,68 +299,66 @@ class Editor extends React.Component {
...
+          </div>
+        </Toolbar>
+
+        <Dropzone accept="image/*, text/*, application/*" onDrop={this.onDrop}>
+          {({ canDrop }) => (
+            <Overlay
+              isOver={canDrop}
+              title={`Drop your file here to import ${canDrop ? '✋' : '✊'}`}
+            >
+              {/*key ensures Carbon's internal language state is updated when it's changed by Dropdown*/}
```
mfix22: @raboid I think we can remove this `key` now since the `Carbon` component doesn't have internal state

mfix22: What do you think?

raboid: Makes sense to me!

PR #699, opened 2019-03-13
Add offline support

mfix22:
Chrome for macOS now supports installing PWAs for desktop. With this method, we can support desktop/offline use cases automatically. 

🤷‍♂️?

Example: https://carbon-ds98yypkc.now.sh/


mfix22:
Getting 504's and 503's from Yarn...

mfix22:
@briandennis should be good to go now

PR #703, opened 2019-03-18
Add gist support to embed

raboid:


components/Editor.js
```diff
@@ -15,6 +15,7 @@ import Carbon from './Carbon'
 import ExportMenu from './ExportMenu'
 import Themes from './Themes'
 import TweetButton from './TweetButton'
+import GistParser from './GistParser'
```
mfix22: Can we call this `GistContainer`? 🙂

raboid:
Okay no code should be displayed until after the `GistContainer` has a chance to load now.

Deployed here: https://carbon-pdcm5r209.now.sh/embed/82d742f4efad9757cc826d20f2a5e5af

pages/embed.js
```diff
@@ -99,6 +101,7 @@ class Embed extends React.Component {
   render() {
     return (
       <Page theme={this.state.theme}>
+        <GistContainer onChange={stateFromGist => this.init(stateFromGist)} />
```
mfix22: Will this still work if we pass the function first class? 
```js
<GistContainer onChange={this.init} />
```

Issue #705, opened 2019-03-19
Fix encoding in Embed links to work nicely with Medium


mfix22:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior: https://github.com/dawnlabs/carbon/issues/549#issuecomment-474411157

**Expected behavior**
Should be able to paste URL from iframe copy, or location from browser bar directly into Medium.

PR #707, opened 2019-03-19
Fix Medium embedding workflow

mfix22:
This encodes the entire URI before writing to `window.location`, requiring that we then decode it when reading from the router. 

Closes https://github.com/dawnlabs/carbon/issues/705

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -63,8 +63,7 @@ class Editor extends React.Component {
   async componentDidMount() {
     const { asPath = '' } = this.props.router
     const { query } = url.parse(decode(asPath), true)
-    const queryParams = getQueryStringState(query)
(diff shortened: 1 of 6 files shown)


mfix22:
Tests are failing since you now need to encode URL's to use the query params as an API, which things like the CLI relies on. 

@briandennis is it worth it to add logic that doesn't decode when the page is initially landed on, but it does after the URL has been updated once? This would enable people to use it as is, but still copy a correctly encoded URL for pasting to Mediumpy.

The other option would be to add a "Copy URL" button that correctly does the encoding then.

mfix22:
@jakedex do you think there is a way we can squeeze in a copy embed URL button instead of copy iframe?

PR #712, opened 2019-03-28
Fix language mime type lookup

gko:
First of all thank you for the project

## Description

When you follow the [example url](https://carbon.now.sh/?bg=rgba(74%2C144%2C226%2C1)&t=night-owl&wt=none&l=typescript&ds=true&dsyoff=20px&dsblur=68px&wc=true&wa=true&pv=56px&ph=56px&ln=true&fm=IBM%20Plex%20Mono&fs=14.5px&lh=141%25&si=false&code=alert(%27test%27)%253B&es=2x&wm=false)
which does contain `l=typescript` you can see that mode is recognized correctly
although CodeMirror doesn't highlight the code. This happens because in place of
a mime type we pass `language` parameter directly(in our case it's «typescript»).

## Solution

I've added a small lookup among known languages to find if we have a mime type
already.

P.S. I also added `.editorconfig` so that indent works properly out of the box.


mfix22:
@gko we use prettier for formatting, so an editor config should not be necessary. Devs can code it however they want, and prettier will format it on commit.

gko:
@mfix22 got it. will revert that commit then.

mfix22:
@gko does this still solve your issue? I want to unify how languages are detect, no matter if it is from `auto` or set by the user

gko:
@mfix22 yep, still works.

PR #715, opened 2019-03-29
update gist handling

raboid:


components/Editor.js
```diff
@@ -62,8 +60,21 @@ class Editor extends React.Component {
...
     const { asPath = '' } = this.props.router
-    const { query } = url.parse(asPath, true)
-    const initialState = getQueryStringState(query)
+    let { state: initialState, gistUri } = getRouteState(asPath)
+
+    if (this.context.gist && gistUri) {
+      try {
+        const gistState = await this.context.gist.get(gistUri)
+
+        initialState = {
```
mfix22: If a gist is found, it should use the gist state as is, not mix it with the localStorage state.

lib/api.js
```diff
@@ -61,10 +61,10 @@ function getGist(uid) {
       const snippet = files[otherFiles[0]]
 
       return {
+        gist: true,
         code: snippet.content,
-        language: snippet.language,
-        owner,
-        config
+        ...(snippet.language ? { language: snippet.language.toLowerCase() } : {}),
+        ...(typeof config === 'object' ? { config } : {})
```
mfix22: I think the ordering here should be: 
```js
{
  ...(typeof config === 'object' ? { config } : {}),
  ...(snippet.language ? { language: snippet.language.toLowerCase() },
  code: snippet.content,
  gist: true
}
```

I also think it was a lot more clear before when we mutated the object, instead of messing around with all these `...`'s (i know I typically like those better 🙂 )

lib/api.js
```diff
@@ -46,7 +46,7 @@ function getGist(uid) {
   return gistClient
     .get(`/gists/${uid}`)
     .then(res => res.data)
-    .then(({ owner, files }) => {
+    .then(({ files }) => {
```
mfix22: @raboid i still want to be able to access the gist_id and owner in the future though, similar to: https://github.com/dawnlabs/carbon/pull/713/files#diff-34b1810563417208c53b912bd37bea3bR19

lib/routing.js
```diff
@@ -83,6 +99,10 @@ export const getQueryStringState = query => {
 }
 
 export const updateQueryString = (router, state) => {
+  if (state.gist) {
```
mfix22: I think this check should happen inside `Editor`, not within this function

lib/routing.js
```diff
@@ -62,7 +65,20 @@ export const deserializeState = serializedState => {
...
 
-const getQueryStringObject = query => {
+export const getRouteState = location => {
+  const { query, pathname } = url.parse(location, true)
+  const state = getQueryStringState(query)
+  const path = escapeHtml(pathname.split('/').pop())
+  const gistUri = path.length >= 19 && path.indexOf('.') === -1 && path
+
+  return {
+    gistUri,
```
mfix22: Is `gistUri` the correct word for this? I am thinking it should be like `gistPath` or `gistId`

pages/embed.js
```diff
@@ -46,23 +45,37 @@ const Page = props => (
...
+
   state = {
     ...DEFAULT_SETTINGS,
     code: DEFAULT_CODE,
     mounted: false,
     readOnly: true
   }
 
-  handleUpdate = updates => {
+  async componentDidMount() {
```
mfix22: Nice 👍

pages/index.js
```diff
@@ -26,7 +26,7 @@ class Index extends React.Component {
       updateQueryString(this.props.router, state)
       saveSettings(
         localStorage,
-        omit(state, ['code', 'backgroundImage', 'backgroundImageSelection', 'filename'])
+        omit(state, ['code', 'backgroundImage', 'backgroundImageSelection', 'filename', 'gist'])
```
mfix22: Do you think it is worth having a separate component state (somewhere) that holds all the gist state? That way the editor doesn't have to omit all of that?

lib/routing.js
```diff
@@ -62,7 +65,20 @@ export const deserializeState = serializedState => {
   return JSON.parse(decodeURIComponent(stateString))
 }
 
-const getQueryStringObject = query => {
+export const getRouteState = location => {
+  const { query, pathname } = url.parse(location, true)
+  const state = getQueryStringState(query)
```
mfix22: Also if `getQueryStringState` is no longer used outside of this function, make sure it's not exported :+1:

components/Editor.js
mfix22: Basically, if it doesn't find a config from the gist, then it can look in localStorage. Everything elsse stays the exact same,

lib/routing.js
```diff
@@ -62,24 +66,18 @@ export const deserializeState = serializedState => {
...
-
-  return state
-}
-
-export const getQueryStringState = query => {
-  const queryParams = getQueryStringObject(query)
-  return Object.keys(queryParams).length ? queryParams : {}
 }
 
 export const updateQueryString = (router, state) => {
```
mfix22: @raboid now that we added `getRouteState`, what do you think about renaming this `updateRouteState`?

raboid: Sounds like a good name to me!

mfix22:
@raboid can you make a test that hits both `/82d742f4efad9757cc826d20f2a5e5af` and `/embed/82d742f4efad9757cc826d20f2a5e5af` and confirms the the URLs don't have a query string, and that the background of the editor is `green`?

PR #727, opened 2019-04-10
support theme highlights in query string

raboid:


components/Editor.js
```diff
@@ -284,7 +282,13 @@ class Editor extends React.Component {
     return (
       <div className="editor">
         <Toolbar>
-          <Themes key={theme} updateTheme={this.updateTheme} theme={theme} />
+          <Themes
+            ref={this.themes}
```
mfix22: What is this `ref` for?

components/Themes/index.js
```diff
@@ -40,50 +42,83 @@ const ThemeItem = ({ children, item, isSelected, onClick }) => (
...
     name: 'Custom Theme',
     selected: null
   }
 
   dropdown = React.createRef()
 
   componentDidMount() {
+    const { queryState } = getRouteState(this.props.router)
+
+    const queryHighlights = queryState
```
mfix22: Since `highlights` are just passed in as a prop now, shouldn't `queryHighlights` be calculated in `Editor` and passed in directly to `<Themes />`

components/Editor.js
raboid: Whoops leftover from a previous approach.

lib/routing.js
```diff
@@ -35,7 +35,19 @@ const mappings = [
   { field: 'wm:watermark', type: 'bool' },
   { field: 'copy', type: 'bool' },
   { field: 'readonly', type: 'bool' },
-  { field: 'id' }
+  { field: 'id' },
+  { field: 'text' },
```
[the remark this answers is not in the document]
mfix22: I also _do not_ want to update the `highlights` query param. I just want to send it as a parameter for Safari.

mixn:
Omg [yes please!](https://github.com/mixn/carbon-now-cli/issues/46) 😍😇🙃☺️🥳🤩

components/Themes/index.js
mfix22: @mixn is passing a single `highlights` query param as JSON sufficient for the CLI? 🙂

components/Editor.js
```diff
@@ -96,11 +96,20 @@ class Editor extends React.Component {
 
   carbonNode = React.createRef()
 
-  updateState = updates =>
-    this.setState(updates, () => !this.gist && this.props.onUpdate(this.state))
+  updateState = updates => {
+    this.setState(updates, () => {
+      // eslint-disable-next-line no-unused-vars
+      const { highlights, ...state } = this.state
```
mfix22: This omission should be handled by the callee

components/Editor.js
```diff
@@ -96,11 +96,20 @@ class Editor extends React.Component {
...
+      !this.gist && this.props.onUpdate(state)
+    })
+  }
 
   updateCode = code => this.updateState({ code })
 
+  updateTheme = theme => this.updateState({ theme })
```
mfix22: Is it possible to just combine `updateTheme` and `updateHighlights` into a single call since they are always run together? 

As a fallback you could even do `updateTheme(theme, highlights)`

mfix22: I think this is going to simplify the rest of the changes as well, so going to hold off reviewing that until this is updated.

components/Themes/index.js
mixn: @mfix22 Could you provide an example of usage via URL (in the deployment below)? That’d be great. 😊

mixn: @mfix22 Tried [this](https://carbon-ek0fn70mc.now.sh/?bg=rgba(171%2C%20184%2C%20195%2C%201)&t=theme%3A45t3iqdbm4b&wt=none&l=auto&ds=true&dsyoff=20px&dsblur=68px&wc=true&wa=true&pv=56px&ph=56px&ln=false&fm=Hack&fs=14px&lh=133%25&si=false&es=2x&wm=false) — unfortunately didn’t work for me? Sorry if I’m missing something.

raboid: @mixn it's expecting the highlight url param to equal something of the form `encodeURIComponent(JSON.stringify({ background: '#fff', text: '#000', ... }))` 

Have you tried something like that?

components/Themes/index.js
```diff
@@ -40,50 +40,71 @@ const ThemeItem = ({ children, item, isSelected, onClick }) => (
...
-    this.setState(({ themes }) => ({
-      preset,
-      highlights: themes.find(({ id }) => id === preset).highlights
-    }))
+  componentDidUpdate(prevProps) {
+    const { isVisible, theme, update } = this.props
+    const { themes } = this.state
+
+    if (prevProps.isVisible && !isVisible) {
+      this.setState({ name: getCustomName(themes) })
```
mfix22: Why does `name` have to be in state? Can it not be derived?

raboid: This sets the initial state for the name which will then be updated as you type into the input.

components/Themes/index.js
raboid: @mixn it should work with a queryString of the form: `?state=${btoa(encodeURIComponent(JSON.stringify({ highlights: { background: 'blue', operator: '#000', text: '#fff', ... } })))}`

components/Themes/index.js
raboid: Hopefully 72a77c8 makes that more clear?

components/Editor.js
```diff
@@ -284,7 +284,7 @@ class Editor extends React.Component {
     return (
       <div className="editor">
         <Toolbar>
-          <Themes key={theme} updateTheme={this.updateTheme} theme={theme} />
+          <Themes key={theme} update={this.updateState} theme={theme} highlights={highlights} />
```
mfix22: @raboid I don't think `key` is necessary anymore, and in fact it will hurt rendering. Is this correct?

components/Themes/index.js
```diff
@@ -40,86 +40,107 @@ const ThemeItem = ({ children, item, isSelected, onClick }) => (
...
   componentDidMount() {
+    const { update, theme, highlights } = this.props
     const storedThemes = getThemes(localStorage) || []
 
     this.setState(({ themes }) => {
       const newThemes = [...storedThemes, ...themes]
 
-      const name = `Custom Theme ${newThemes.filter(({ name }) => name.startsWith('Custom Theme'))
-        .length + 1}`
+      this.selectedTheme = newThemes.find(({ id }) => id === theme) || DEFAULT_THEME
```
mfix22: Should this line and the following `update` be moved into the callback in the second `setState` argument?

components/Themes/index.js
mfix22: I don't care either way 🙂feel free to keep this or revert. If you want it to be `input`, I'd prefer it be called input everwhere and not renamed to `name` in _some_ places.

components/Themes/index.js
mfix22: Hey @mixn yeah it's not ready quite yet, I was more asking from an API standpoint. I'll let you know soon.

mfix22: Hey @mixn yeah it's not ready quite yet, I was more asking from an API standpoint. I'll let you know soon.

components/Editor.js
raboid: Yeah good catch 👍

components/Themes/index.js
mfix22: Hey @mixn yeah it's not ready quite yet, I was more asking from an API standpoint. I'll let you know soon.

components/Themes/index.js
mfix22: I don't care either way 🙂feel free to keep this or revert. If you want it to be `input`, I'd prefer it be called input everwhere and not renamed to `name` in _some_ places.

components/Editor.js
```diff
@@ -81,6 +83,25 @@ class Editor extends React.Component {
       loading: false
     }
 
+    const storedThemes = getThemes(localStorage)
+
+    newState.themes = storedThemes
+      ? [...storedThemes, ...this.state.themes]
```
mfix22: I believe you can just do `[...storedThemes, ...this.state.themes]` without the ternary 👍

components/Themes/index.js
```diff
@@ -40,131 +47,58 @@ const ThemeItem = ({ children, item, isSelected, onClick }) => (
...
-  }
-
   componentDidUpdate(prevProps) {
-    const { isVisible, theme, update } = this.props
-    const { themes } = this.state
-
-    if (prevProps.isVisible && !isVisible) {
-      this.setState({ input: getCustomName(themes) })
-      update({ highlights: themes.find(({ id }) => id === theme).highlights })
+    if (prevProps.isVisible && !this.props.isVisible) {
```
mfix22: Can we just convert this to a `getDerivedStateFromProps`

components/Editor.js
```diff
@@ -83,7 +83,7 @@ class Editor extends React.Component {
       loading: false
     }
 
-    const storedThemes = getThemes(localStorage)
+    const storedThemes = getThemes(localStorage) || []
```
mfix22: Dang! always thought `[...null, ...[]]` was fine, but I guess not!

components/Themes/index.js
mixn: @mixn Thanks for letting me know. :) It definitely sounds sufficient. Would be amazing if it’d be optional as the others and would behave the same way. I prepare the URL [here](https://github.com/mixn/carbon-now-cli/blob/master/cli.js#L193) after it has gone through all previous steps, it’s really, really basic. Initially it’s based on [the default settings](https://github.com/mixn/carbon-now-cli/blob/master/src/helpers/default-settings.js).

The only small problem I see is it adding to [this issue](https://github.com/mixn/carbon-now-cli/issues/44) / the length of URLs. But that’s a separate thing. :)

PR #730, opened 2019-04-13
Added functionality to adapt window title color

muhammadmuzzammil1998:
# The problem

Titles work fine in dark themes but when the theme is light, the color makes it difficult to read title sometimes.
For example:
This is an image using "Solarized (Dark)" theme.
image
And this is an image using "Solarized (Light)" theme.
image

The title on the dark theme is fine but on the light theme, it's barely visible.

# A solution

This pull request adds a function to check if the background of window is dark or light and applies that function to change color of the window title.

These images are from a local version running with this fix:
carbon (51)
carbon (50)
carbon (49)
carbon (48)
carbon (47)


muhammadmuzzammil1998:
Thanks, @mfix22!

mfix22:
@muhammadmuzzammil1998 thank you 🙏! Very helpful.

PR #734, opened 2019-04-16
Custom font

raboid:


components/Editor.js
```diff
@@ -86,6 +87,10 @@ class Editor extends React.Component {
       newState.language = unescapeHtml(newState.language)
     }
 
+    if (newState.fontFamily && !FONTS.find(({ id }) => id === newState.fontFamily)) {
```
mfix22: @raboid I don't think this is where fallback logic should happen. That's why I moved all of those other fallbacks for theme/highlight out of cDM (newState.language shouldn't even be here).

If you want to prevent an invalid `fontFamily` from being passed, that logic should happen deeper in the component tree, even in `Editor.render`.

PR #739, opened 2019-04-19
Add Japanese translation of README

uraway:
I have translated README into Japanese. I would be happy if this could help to let the Japanese know this great project even just a little.


mfix22:
Thanks so much @uraway!!!

PR #762, opened 2019-05-16
Introducing Applitools

megsachdev:
Hello!
At times maintaining separate visual test cases becomes a tiring task so using your existing examples setup we tried coming up with something that can ease this process. So I chose to use Cypress as the browser test harness, which Applitools has an SDK for. I started off with writing tests for some of the components so it's sort of a "smoke test" but you can add more if you'd like.

Visual regression involves comparing the pixels visually like a human would, ideally ignoring differences that a human couldn't perceive e.g. 1 pixel changes, differences in how browsers render images, drop shadows, etc. to avoid false positives; but you can still do a strict 1:1 pixel match if that's your preference.

Applitools thankfully  [provides an SDK](https://applitools.com/tutorials/cypress.html)  that made it trivial to write tests using Cypress. So other than the things I included what you'll need to do is  [create your own Applitools account](https://applitools.com/users/register)  and  [then setup your GitHub CI integration](https://applitools.com/docs/topics/integrations/github-integration.html)  if you want it to run automatically on Pull Requests; optional, but probably a good idea.
Here is the quick reference to Applitools: https://applitools.com/

Every test we write will include selectors for elements. To save a lot of headaches, we should write selectors that are resilient to changes.

Oftentimes we see users run into problems targeting their elements because:
Your application may use dynamic classes or ID’s that change
Your selectors break from development changes to CSS styles or JS behavior

Luckily, it is very easy to avoid both of these problems:
Don’t target elements based on CSS attributes such as: id, class, tag
Don’t target elements that may change their textContent
Add data-* attributes to make it easy to target elements

That's what I have done here, I added data attributes to the existing components to make the code neat and consistent.

To run Applitools:

Create Applitools account
Go to the menu in the top right corner and click the profile icon on the far right.
image

Click on "My API Key" and copy your key
Go to your terminal and paste
`export APPLITOOLS_API_KEY=Your_API_Key_Here`
Go to your directory and run:
`yarn test:e2e`
In your Applitools dashboard, you should see the results of your tests
step0004

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -5,6 +5,7 @@ import { COLORS } from '../lib/constants'
 
 const Button = ({
   id,
+  cypressElementId,
   onClick = () => {},
   className = '',
   background = COLORS.BLACK,
@@ -23,7 +24,14 @@ const Button = ({
   margin = 0,
   title
 }) => (
-  <button id={id} onClick={onClick} className={className} disabled={disabled} style={style}>
+  <button
+    id={id}
+    onClick={onClick}
+    className={className}
+    disabled={disabled}
+    style={style}
+    data-cy={cypressElementId}
+  >
     {title && <VisuallyHidden>{title}</VisuallyHidden>}
     {children}
     <style jsx>
--- a/components/Button.js
+++ b/components/Button.js
@@ -5,7 +5,7 @@ import { COLORS } from '../lib/constants'
 
 const Button = ({
   id,
-  cypressElementId,
+  'data-cy': cypressElementId,
   onClick = () => {},
   className = '',
   background = COLORS.BLACK,
--- a/cypress/integration/visual-testing/button-test.spec.js
+++ b/cypress/integration/visual-testing/button-test.spec.js
@@ -1,0 +1,48 @@
+/* global cy,before,after */
+import { environment } from '../../util'
+describe('Visual Regression Testing', () => {
+  before(() => {
+    cy.eyesOpen({
+      appName: 'Carbon',
+      testName: 'Button',
+      browser: environment
+    })
+    cy.visit('http://localhost:3000')
+  })
+  after(() => {
+    cy.eyesClose()
+  })
+  beforeEach(() => {
+    cy.reload()
+  })
+
+  it('test export button', () => {
+    cy.get('[data-cy=export-button]').click()
+    cy.eyesCheckWindow({
+      tag: 'export button',
+      sizeMode: 'selector', //mode
+      selector: '.page'
+    })
+  })
+
+  it('test display button', () => {
+    cy.get('[data-cy=display]').click()
+    cy.eyesCheckWindow({
+      tag: 'display button',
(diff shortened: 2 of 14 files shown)


cypress/integration/visual-testing/button-test.spec.js
```diff
@@ -0,0 +1,48 @@
+/* global cy,before,after */
+import { environment } from '../../util'
+describe('Visual Regression Testing', () => {
+  before(() => {
+    cy.eyesOpen({
+      appName: 'Carbon',
+      testName: 'Button',
+      browser: environment
+    })
+    cy.visit('http://localhost:3000')
```
mfix22: Thanks for this! For each of the `visit`'s do you need to prefix everything with `http://localhost:3000` or can you just use `cy.visit('/')`.

That way it will work with Cypress's `--baseUrl` flag

components/Button.js
```diff
@@ -23,7 +24,14 @@ const Button = ({
...
   title
 }) => (
-  <button id={id} onClick={onClick} className={className} disabled={disabled} style={style}>
+  <button
+    id={id}
+    onClick={onClick}
+    className={className}
+    disabled={disabled}
+    style={style}
+    data-cy={cypressElementId}
```
mfix22: Is it possible to just pass `Button` `data-cy` directly instead of renaming the prop `cypressElementId`? 🙂

cypress/integration/visual-testing/button-test.spec.js
megsachdev: I could have used the URL mentioned the cypress.json but that's pointing to production. So I was unable to use that.

components/Button.js
megsachdev: Sure I can change that.

cypress/integration/visual-testing/button-test.spec.js
mfix22: Can you no have `test:e2e` use `--baseUrl=http://localhost:3000`?

Or even just run it directly as `yarn test:e2e --baseUrl=http://localhost:3000`

I really appreciate you doing this work!

components/Button.js
megsachdev: done :)

cypress/integration/visual-testing/button-test.spec.js
megsachdev: I can run with these commands on my local but how will Travis pick this?
Do you want me to add another npm script for visual tests?

mfix22:
Thanks so much for your help Megha! 🙏

megsachdev:
Thank you so much for the approval. Are we good to merge this?

mfix22:
@megsachdev I'm just doing a little more testing and then I am going to merge! Thanks for all your work 👌

mfix22:
@all-contributors please add @megsachdev for code and tests

megsachdev:
Thank you so much @mfix22..

PR #786, opened 2019-06-12
Fix Firefox open export

### LINKED ISSUE #785: Open image doesn't work in Firefox
**Describe the bug**
When "Open" is selected from the menu, the image disappears after a moment.

**To Reproduce**
1. Open firefox.
2. Type a code sample
3. Click Export -> Open

The image opens for a short period, but almost immediately redirects back to the main carbon.sh page, giving no opportunity to copy it.

**Expected behavior**
A `blob:` version of the image opens and stays open, like:

blob:https://carbon.now.sh/7ddb0b86-5dec-4ec2-955b-a9521597694f

On Chrome it works fine.

- OS: Linux
- Firefox 67.0.1

mfix22:
Closes #785

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -97,6 +97,10 @@ class Editor extends React.Component {
       window.navigator &&
       window.navigator.userAgent.indexOf('Safari') !== -1 &&
       window.navigator.userAgent.indexOf('Chrome') === -1
+    this.isFirefox =
+      window.navigator &&
+      window.navigator.userAgent.indexOf('Firefox') !== -1 &&
+      window.navigator.userAgent.indexOf('Chrome') === -1
   }
 
   carbonNode = React.createRef()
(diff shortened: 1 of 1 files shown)


jakedex:
Nice! 👍

travisdowns:
Thanks @mfix22 - that has to be among the quickest turnarounds I've seen after filing an issue (if the issue is acknowledged at all).

How do I know when the fix has been "deployed", i.e., is live on the public site?

mfix22:
Thank you for saying so! 
I will post a new release in the [releases](https://github.com/dawnlabs/carbon/releases) page and it will include this PR 🙂

travisdowns:
Cool, I'll check the fix then.

Thanks again.

travisdowns:
I tried the fix at:

https://carbon-ga8pm71x3.now.sh

and it doesn't work, although the behavior is different: the screen just flashes for a second (I don't even see the image) and stays on the editor.

However, the main carbon.now.sh seems now to be working fine!

PR #807, opened 2019-07-09
nef - integration with Carbon

miguelangel-dev:
## Description
We've added an integration with Carbon in our tool `nef` - you can read more about it [here](https://nef.bow-swift.io/docs/tutorials/first-steps-with-carbon/)

## What is nef?
In a general way, `nef` is a CLI that lets you have compile-time verification of your documentation written as **Xcode Playground**. Now, from version [0.3.0](https://www.47deg.com/blog/nef-0-3-0-release/#carbon-integration-2) we have integrated Carbon, so you can write in **official Apple playgrounds**, and export the whole snippets with just one command.

## Request
We love Carbon, and we usually use it; this is the reason for this integration and the request for adding in the Carbon's tool list.


miguelangel-dev:
@mfix22 LGTM!

Following your suggestions, I have updated the link to `nef` in `README` - now it is pointing to Carbon [section](https://github.com/bow-swift/nef#-exporting-carbon-code-snippets)

> Note, it contains an expandable section with more details

PR #835, opened 2019-07-31
Add Framer X theme

davo:
Adding support for Framer X theme, loosely inspired from their [API documentation site](https://www.framer.com/api/frame/). 

carbon (2)

## Description

Was not possible to replicate 100% of the properties from their theme, this is a very good approximation.

Screen Shot 2019-07-31 at 1 13 35 PM


mfix22:
@davo since we now support saving your custom themes for later, we don't want to add new custom themes that are simply new configurations of highlights. Is there something preventing you from saving this theme for yourself? 

We are soon going to add a way to share custom themes easier as well 👍

davo:
@mfix22 thanks for the clarification. Nothing is preventing me from using the custom theme feature but I thought I would be a nice addition because of the following reasons:

1. The custom theme feature flattens some attributes [e.g.: .cm-variable-2, .cm-variable-3 gets overridden by the inline style variable]. So it does not reflect the actual theme customizations I wanted to achieve.
2. Really wanted to share this with the Framer X community, in order to improve their documentation and tutorials.

If this is the direction moving forward, shouldn't be a good idea to update the [contributing](https://github.com/dawnlabs/carbon/blob/master/.github/CONTRIBUTING.md) readme? It was my guiding principle to make the contribution as clean and simple as possible.

Let me know your thoughts and thank you for making this awesome tool.

mfix22:
@davo you're right, CONTRIBUTING.md should have been updated. We wanted to accept people adding their own themes until they were able to create their own.

Is it possible to add the "Framer X" theme without a custom `CSS` file? If that is possible, I would _consider_ adding it since our contributing guidelines weren't clear. 

What is causing this?
> [e.g.: .cm-variable-2, .cm-variable-3 gets overridden by the inline style variable]

mfix22:
Hi @davo, sorry for the delay here. 

Although you have shared a truly beautiful theme,  unfortunately it does not make sense to add it to Carbon at this time. As of Carbon `4.0.0`, the core Carbon team has not been accepting new themes, and instead are encouraging people to use custom themes and presets. 

Hopefully in the future we could allow sharing themes like this one through something like a theme marketplace.

PR #836, opened 2019-08-02
Store Unsplash background image not dataturl

### LINKED ISSUE #832: Error on background image of preset
First of all, i thank you all for this work of design in code :smiley:

**Describe the bug**
Changing a stored preset containing background image gives an unexpected error.

**To Reproduce**

1. In a fresh new, use a background image
2. Save as preset
3. Change any setting, then save as new preset
3. See error

**Expected behavior**
Saves two presets with same background image.

**Info:**

- OS: Windows
- Browser: Latest Chrome

mfix22:
Closes #832


mfix22:
@GuiDevloper can you visit https://carbon-git-store-background-image-not-dataturl.carbon-app.now.sh and try saving presets? I updated it to store the Unsplash image and not the selection.

This means that when you load the preset though, you won't be able to adjust the selection. Does this work for you?

GuiDevloper:
The deploy works good, @mfix22.
The main here is to avoid mysterious bugs.

I log `localStorage.CARBON_PRESETS` and see decreased presets becoming almost limitless.
But too a key called `icon` still stores the base64, if possible saving URL on this one will create a _"limitless storage"_ (in human terms).

Hey, me and future users thank for the attention and availability!

PS: I opened the last [current deploy](https://carbon-nhgd7pcx5.now.sh/) and see that it is not loading Unsplash, sorry if is intentional!

PR #849, opened 2019-08-17
Add support for first line number setting

andrewda:
image

## Description

This adds a setting (when `Line numbers` is on) that allows the user to specify the first line number.

Closes https://github.com/carbon-app/carbon/issues/842


components/Settings.js
```diff
@@ -79,6 +81,16 @@ const WindowSettings = React.memo(
           enabled={lineNumbers}
           onChange={onChange.bind(null, 'lineNumbers')}
         />
+        {lineNumbers && (
+          <Text
+            label="First line number"
+            type="int"
+            value={firstLineNumber}
+            minValue={1}
```
mfix22: Let's do `0` in case someone wants zero-based indexing 🤷‍♂

lib/routing.js
```diff
@@ -56,6 +56,7 @@ const readMappings = [
   { field: 'pv:paddingVertical' },
   { field: 'ph:paddingHorizontal' },
   { field: 'ln:lineNumbers', type: 'bool' },
+  { field: 'fl:firstLineNumber' },
```
mfix22: Let's add a type that checks if the value is a number, or returns `undefined`

components/Text.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
```
mfix22: I'm going to touch this up a bit. Thanks for creating this though!

andrewda: Looks much better now, thanks for the touch-ups!

mfix22:
Thanks again Andrew! 🙏

PR #854, opened 2019-08-30
Update FUNDING.yml with Ethereum address

### LINKED ISSUE #853: Add Ethereum address to FUNDING.yml
I'd like to see an Ethereum address added to the `custom` property of the `FUNDING.yml` file for GitHub Sponsors. This would allow a canonical address to send funds to. For example, I'm building a project that allows developers to setup one recurring donation that is split between all of their dependencies with an Ethereum address: https://sustainus.io

lastmjs:
## Description

Adding an Ethereum address per https://github.com/carbon-app/carbon/issues/853#issuecomment-526777956

Closes #853

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -7,4 +7,4 @@ community_bridge: # Replace with a single Community Bridge project-name e.g., cl
 liberapay: # Replace with a single Liberapay username
 issuehunt: # Replace with a single IssueHunt username
 otechie: # Replace with a single Otechie username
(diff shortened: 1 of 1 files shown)


lastmjs:
Yes, I could link it to Etherscan if you'd like, that would give a generic overview of the address

mfix22:
@lastmjs sure. Whatever is the most user-friendly URL.

lastmjs:
Etherscan I believe is the most used block explorer for Ethereum, so for crypto users it should be pretty nice. The best thing to do would probably be to create a very user-friendly landing page at something like sustainus.io/address/whatever-ethereum-address, that explains that donations are accepted to this address and gives the user options for which crypto tools to use to donate...I do not have that created yet

PR #858, opened 2019-09-11
Fix translation error

Merlintor:
I only changed three words in the german README, because it wasn't valid german before.


mfix22:
Thank you so much @Merlintor! Just want to check in with the original author of the German README and then we can merge. 

@mixn is this still good-to-go from your perspective? 🙂

mfix22:
@all-contributors add @Merlintor for translation

PR #868, opened 2019-09-22
Cypress parallel

mfix22:


mfix22:
No real speed up from `--parallel` flag. Closing...

PR #878, opened 2019-10-06
Load selected lines from query params

### LINKED ISSUE #843: Load selected lines from query params
See https://github.com/carbon-app/carbon/issues/824#issuecomment-513694282 for context

- [x] Should use existing state/routing system

nemesv:
Closes #843

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -408,11 +408,13 @@ function useGutterClickHandler(props, editorRef) {
   })
 
   React.useEffect(() => {
-    if (editorRef.current) {
+    if (editorRef.current && Object.keys(state.selected).length > 0) {
       editorRef.current.editor.display.view.forEach((line, i) => {
-        if (line.text && line.gutter) {
-          line.text.style.opacity = state.selected[i] === false ? 0.5 : 1
-          line.gutter.style.opacity = state.selected[i] === false ? 0.5 : 1
+        if (line.text) {
+          line.text.style.opacity = state.selected[i] === true ? 1 : 0.5
+        }
+        if (line.gutter) {
+          line.gutter.style.opacity = state.selected[i] === true ? 1 : 0.5
         }
       })
     }
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -39,7 +39,6 @@ class Carbon extends React.PureComponent {
     onGutterClick: noop
   }
   state = {}
-  editorRef = React.createRef()
 
   handleLanguageChange = debounce(
     (newCode, language) => {
@@ -101,7 +100,7 @@ class Carbon extends React.PureComponent {
 
   onMouseUp = () => {
     if (this.currentSelection) {
-      const { editor } = this.editorRef.current
+      const { editor } = this.props.editorRef.current
       const startPos = editor.charCoords(this.currentSelection.from, 'window')
       const endPos = editor.charCoords(this.currentSelection.to, 'window')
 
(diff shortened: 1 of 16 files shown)


lib/routing.js
```diff
@@ -72,7 +72,8 @@ const readMappings = [
   { field: 'readonly', type: 'bool' },
   { field: 'id' },
   { field: 'highlights', type: 'parse' },
-  { field: 'code', type: 'decode' }
+  { field: 'code', type: 'decode' },
+  { field: 'selectedLines' }
```
mfix22: Can se do `sl:selectedLines`?

components/Carbon.js
```diff
@@ -377,13 +376,21 @@ function useHighlightLoader() {
...
+  { type, lineNumber, numLines, selectedLines }
+) {
   const newState = {}
 
   if (type === 'GROUP' && prevLine) {
     for (let i = Math.min(prevLine, lineNumber); i < Math.max(prevLine, lineNumber) + 1; i++) {
       newState[i] = selected[prevLine]
     }
+  }
+  if (type === 'MULTILINE') {
```
mfix22: Hmm what is `MULTILINE` used for? Just for selecting  any arbitrary list of lines? 🙂

components/Carbon.js
```diff
@@ -402,27 +409,34 @@ function selectedLinesReducer({ prevLine, selected }, { type, lineNumber, numLin
...
-  }, [state.selected, props.children, props.config])
+  }, [state.selected, props.children, props.config, editorRef])
 
-  return React.useCallback(function onGutterClick(editor, lineNumber, gutter, e) {
-    editorRef.current = editor
+  React.useEffect(() => {
+    let selectedLines = (props.config.selectedLines || '')
+      .split(',')
+      .map(i => parseInt(i))
+      .filter(i => !isNaN(i))
```
mfix22: This number filtering/mapping should happen during the query param parsing here: https://github.com/carbon-app/carbon/pull/878/files#diff-7bd084e1642aa17e0074c18f5a25a0aeR76

components/Carbon.js
nemesv: Yes, that was my idea to support selecting a list of lines. Do you have a better suggestion for the name? In theory the other two actions GROUP and LINE could be also expressed as MULTILINE but I didn't want to change to much.

components/Carbon.js
nemesv: sure, I will move it

lib/routing.js
nemesv: Ok, I will rename it

components/Carbon.js
```diff
@@ -402,27 +409,31 @@ function selectedLinesReducer({ prevLine, selected }, { type, lineNumber, numLin
   }
 }
 
-function useGutterClickHandler(props) {
-  const editorRef = React.useRef(null)
+function useSelectedLines(props, editorRef) {
   const [state, dispatch] = React.useReducer(selectedLinesReducer, {
     prevLine: null,
     selected: {}
   })
```
mfix22: @nemesv quicks thoughts. Now that we are also story `selectedLines` in the `Editor` state, we can probably remove this local state and just call something like `this.props.setSelectedLines` instead. That way we don't have this state in two places. 

We _don't_ need it in this PR, but do you think that sounds feasible?

mfix22:
@nemesv this is looking really good! I will test it out more tomorrow and try and get it merged.

mfix22:
@nemesv sorry about this, but can you fix the merge conflicts? Feel free to just check out `master` and then re add yourself as a contributor.

Alternatively, I will add you as a contributor once we merge this automatically!

mfix22:
@nemesv this is awesome! Great work! Really appreciate your help 👍

PR #879, opened 2019-10-07
Display whitespace

### LINKED ISSUE #877: Display whitespace
I was making some temporary screenshots from my editor while out of wifi range and discovered a starling difference. My editor shows white space, including vertical lines.

Generally you might think that's not super useful, but it is VERY useful when dealing with JSON, and writing say a JSON Schema.

It would be great if such an option could be added.

<3

baktiaditya:
Closes #877

Screen Shot 2019-10-08 at 06 48 14

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -47,14 +47,18 @@ class Carbon extends React.PureComponent {
         // try to set the language
         const detectedLanguage = hljs.highlightAuto(newCode).language
         const languageMode = searchLanguage(detectedLanguage)
+
         if (languageMode) {
           return languageMode.mime || languageMode.mode
         }
       }
+
       const languageMode = searchLanguage(language)
+
       if (languageMode) {
         return languageMode.mime || languageMode.mode
       }
+
       return language
     },
     ms('300ms'),
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -33,6 +33,7 @@ function searchLanguage(l) {
 }
 
 function noop() {}
+
 class Carbon extends React.PureComponent {
   static defaultProps = {
     onChange: noop,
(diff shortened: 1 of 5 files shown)


components/Carbon.js
```diff
@@ -377,6 +379,16 @@ function useHighlightLoader() {
   }, [])
 }
 
+let showInvisibleLoaded = false
+function useShowInvisibleLoader() {
```
mfix22: Let's adjust this to use `import()` instead, and only call `import()` if `whiteSpace` is true

mfix22: There is probably no need to check if it is already loaded in this case

components/Settings.js
```diff
@@ -95,6 +96,11 @@ function WindowSettings({
           />
         </div>
       )}
+      <Toggle
+        label="Whitespace"
```
mfix22: You can leave the variable as whitespace, but can you change the `label` to `"Hidden characters"`?

components/Carbon.js
baktiaditya: Sorry @mfix22 i don't know how to do a dynamic import based on boolean value 😢

baktiaditya: @mfix22 sorry, i don't know how to do dynamic import properly 😞

baktiaditya: many attempts failed

mfix22:
@baktiaditya can you rebase it on `master` so that only your changes are present in this PR? Sorry about all the hassle! I don't have time to do it myself at the moment

mfix22:
@baktiaditya what if instead of using cm-show-invisibles, we just replaced the whitespace characters in `this.props.children`? 

Something like: 
```js
this.props.children
  .replace(/(\r\n|\r|\n)/g, '¬\n')
  .replace(/ /g, '·')
```

Thoughts?

mfix22:
@all-contributors add @baktiaditya for code

Issue #880, opened 2019-10-08
Add manual config for image width


mearns:
I'd like to be able to specify a fixed width for the image. This is particularly important when creating a series of image for an article; aesthetically, I'd like them to all be the same width.

### Expected Behavior

When auto-adjust width is not enabled, I expect a settings control to change the width of the image.

### Actual Behavior

There is no option to manually control the width of the image, only to make it auto-adjust or an unspecified fixed width.

mearns:
Related to #1

mfix22:
@all-contributors add @mearns for ideas

mfix22:
@all-contributors add @mearns for ideas

PR #881, opened 2019-10-08
Implement manual selection for minimum width

mearns:
Adds a new setting to manually configure the minimum width of the image if auto-adjust is not enabled. At least partially addresses #880, though it doesn't currently set the exact width of the "terminal window", it sets the minimum width.


mearns:
Making the window wide was actually only tangentially related to the selectable width. I definitely understand that it doesn't look right, the reason was that as you adjust the width in the settings menu, it moves the settings menu. This wouldn't be a problem if it was implemented with a drag handle. I'll see if I can get some time to look into that.

PR #887, opened 2019-10-15
Add a11y-dark theme

ericwbailey:
This PR adds my [a11y-dark theme](https://github.com/ericwbailey/a11y-syntax-highlighting#a11y-dark). It is a syntax highlighting theme that uses all WCAG AAA compliant values for color contrast.

Code in Carbon highlighted using the a11y-dark theme.

Thanks for considering it! Carbon's a really great app, and I'm excited to be able to contribute 🎉

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -50,6 +50,25 @@ export const THEMES = [
     }
   },
   {
+    id: 'a11y-dark',
+    name: 'a11y-dark',
+    highlights: {
+      background: '#2b2b2b',
+      text: '#f8f8f2',
+      variable: '#00e0e0',
+      attribute: '#abe338',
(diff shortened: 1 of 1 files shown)


lib/constants.js
```diff
@@ -49,6 +49,25 @@ export const THEMES = [
       meta: '#555'
     }
   },
+  {
+    id: 'a11y-dark',
+    name: 'a11y-dark',
```
mfix22: Can we name it `A11y Dark`? `id` can stay as is.

ericwbailey: Happily!

mfix22:
@all-contributors add @ericwbailey for code

PR #895, opened 2019-10-23
nef - Xcode Editor Extension

miguelangel-dev:
Repository: https://github.com/bow-swift/nef-plugin

This project provides an extension for Xcode to integrate some nef features directly in the IDE. Using the core of nef, you can export snippets from your code selection directly in Xcode.

nef-plugin-action-export
Screenshot 2019-10-23 at 15 52 00

**Note** I think it could replace the old [Xcode extension](https://github.com/StevenMagdy/CarboNow4Xcode), but I let you to consideration ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Check out these projects our awesome community has created:
 - [Vim `carbon-now-sh`](https://github.com/kristijanhusak/vim-carbon-now-sh) - Open up the selection in your current Vim/Neovim using function `CarbonNowSh()`
 - [Emacs `carbon-now-sh`](https://github.com/veelenga/carbon-now-sh.el) - Open up the selection in your current Emacs using interactive function `carbon-now-sh`
 - [Xcode `carbon-now-sh`](https://github.com/StevenMagdy/CarboNow4Xcode) - Open up your current selection in `carbon.now.sh`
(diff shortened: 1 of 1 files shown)


mfix22:
@miguelangel-dev what is the difference between this and the plugin we already have included on the README: https://github.com/carbon-app/carbon/blob/b5b9604e2ab4704067fff4b2b2b1c93a2faa5fbf/README.md#tools

miguelangel-dev:
> @miguelangel-dev what is the difference between this and the plugin we already have included on the README: https://github.com/carbon-app/carbon/blob/b5b9604e2ab4704067fff4b2b2b1c93a2faa5fbf/README.md#tools

Ey @mfix22 !

Thanks for your response, and it is a great question. 

`nef` is a set of tool for Xcode Playgrounds, and between other things let you from CLI export block of code into carbon-snippets. On another hand, we have this new product `nef Xcode Extension` taking advantage of `nef` let the user use carbon export snippets directly in Xcode. Finally, the main difference with the current [Xcode plugin] is that `nef` does not present to the user a webview with carbon page; it automatically download the snippet without any other user interaction :) - it is suuper cool and integrate perfectly carbon-Xcode thanks to nef.

mfix22:
@miguelangel-dev thanks for clarifying! Should we remove the old Nef link then?

miguelangel-dev:
> @miguelangel-dev thanks for clarifying! Should we remove the old Nef link then?

IMO both are interesting, because you can use nef from CLI and directly into Xxode depends on your preferences, even automate the process (getting carbon images) in CI using nef (tool); they are two products although one depends on another one.

PR #900, opened 2019-10-25
Add escape listener to selection editor

mfix22:


mfix22:
@arturhun does this solve your issue, or would you still need a setting to disabled the Selection editor altogether?

arturhun:
@mfix22 it's of course a good thing to start with. However, if you use the editor for a while, you'll notice how much the formatter gets in the way - basically any selection (via mouse-drag or double-click) results in the formatter showing up. Personally, I have never needed to format text, but I constantly need to select and rearrange code.

That's why I would recommend an opt-in approach for this feature: dedicated setting "Enable formatting" or maybe a tool icon.

Thanks.

mfix22:
@arturhun awesome, thanks for the feedback. I think you are correct that it is distracting.

PR #906, opened 2019-11-05
Add cobalt2 theme

b3u:
Added the cobalt2 theme by @wesbos

It's based off of the vs-code version of the theme. I tried to stay true to it, but the limitations of the Code Mirror classes made it difficult. It looks nice to me, even if it doesn't look exactly like the original. (Not all versions of the theme look the same; Prism, atom, vs-code, sublime text, etc.)

carbon


mfix22:
@b3u we already have a Cobalt theme, which was also based on Wes' theme. 

If you have an issue with our current Cobalt theme, please open an issue or PR for consideration that updates that theme, as we won't support both.

PR #909, opened 2019-11-26
Display login button on all pages

andrewda:
## Description

This adds the Login/Account button to the top of all pages, including the account and snippets page. This makes transitions between pages look more natural as the button doesn't suddenly disappear. In addition, it adds the previously missing ability to sign out on the account/snippets page, which might be useful if the user discovers they're using the wrong account.

## Screenshots

|Before|After|
|------|-----|
|image|image|

### DIFF
--- a/components/LoginButton.js
+++ b/components/LoginButton.js
@@ -87,7 +87,7 @@ function LoginButton({ isVisible, toggleVisibility }) {
             height: 100%;
           }
           div :global(.profile-button) {
-            max-width: 142px;
+            max-width: 218px;
             min-height: 40px;
           }
           span {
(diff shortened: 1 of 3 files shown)


components/LoginButton.js
```diff
@@ -87,7 +87,7 @@ function LoginButton({ isVisible, toggleVisibility }) {
             height: 100%;
           }
           div :global(.profile-button) {
-            max-width: 142px;
+            max-width: 218px;
```
mfix22: @andrewda what is the reason behind this change?

andrewda: Ah good catch, didn't realize I committed that change. My name was cut off to "Andre..." and I was playing with increasing the size a little to allow longer names (this was enough to fit my full name with some room to spare, without any real impact on the UI).

If you'd prefer keep it at 142 though I can definitely revert that change! 👍

mfix22: @andrewda no worries! A little longer is great! Would you mind posting a screenshot?

andrewda: Sure!

## Before

image

## After

image

PR #916, opened 2019-12-09
Add support for aspect ratios, multiple file, emojies and background image spanning across files

maximgladkov:


lib/constants.js
```diff
@@ -1,5 +1,13 @@
 import toHash from 'tohash'
 
+export const DEFAULT_ASPECT_RATIO = 1
+export const ASPECT_RATIOS = [
+  { id: 'auto', name: 'Auto', value: null },
+  { id: '1:1', name: '1:1', value: 1 },
+  { id: '4:5', name: '4:5', value: 4 / 5 },
+  { id: '1.91:1', name: '1.91:1', value: 1.91 / 1 }
```
mfix22: What is `1.91/1`?

lib/constants.js
```diff
@@ -1,5 +1,13 @@
 import toHash from 'tohash'
 
+export const DEFAULT_ASPECT_RATIO = 1
```
mfix22: The default should be how Carbon production currently is, which if I am not mistake is "Auto" in this case

components/style/Reset.js
```diff
@@ -237,6 +237,10 @@ export default () => (
         text-transform: capitalize;
       }
 
+      .uppercase {
```
mfix22: What is this being used for?

components/Editor.js
```diff
@@ -135,28 +138,43 @@ class Editor extends React.Component {
...
+    })
+
     const width = node.offsetWidth * exportSize
     const height = squared ? node.offsetWidth * exportSize : node.offsetHeight * exportSize
 
     const config = {
       style: {
         transform: `scale(${exportSize})`,
-        'transform-origin': 'center',
+        'transform-origin': '0% 0%',
```
mfix22: Why is this being changed? Was this required?

components/Editor.js
```diff
@@ -135,28 +138,43 @@ class Editor extends React.Component {
...
-          map.set(node, node.innerHTML)
+          map.set(node, { innerHTML: node.innerHTML })
           node.innerText.match(/%[A-Fa-f0-9]{2}/g).forEach(t => {
             node.innerText = node.innerText.replace(t, encodeURIComponent(t))
           })
         }
       })
     }
 
+    node.querySelectorAll('.slider').forEach(node => {
```
mfix22: What is this code doing?

components/Button.js
```diff
@@ -20,6 +20,7 @@ const Button = ({
   flex = 1,
   padding = 0,
   margin = 0,
+  width = 'auto',
```
mfix22: Do we need to make this change? I would prefer not to…

components/Box.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
```
mfix22: I don't think we should create a "Box" component. If we end up using it, `Box` will just become overloaded to the point of it being equally useful as `div`.

If you want to use default values for margins, use the CSS variables found in: `components/style/Typography.js`

components/Editor.js
```diff
@@ -425,6 +449,14 @@ class Editor extends React.Component {
             </Overlay>
           )}
         </Dropzone>
+        <Box marginTop="1rem" display="flex" justifyContent="center">
+          <Toolbar>
+            <AspectRatioButtons
```
mfix22: I think this toolbar needs to be in the settings, under "Settings (Gear Icon) > Window"

mfix22:
Not sure what the goal of this PR is. Please reopen once the feedback is addressed 🙂

PR #917, opened 2019-12-10
Update readme as the english's header version

felipefialho:
Before

Screen Shot 2019-12-09 at 22 35 03

After

Screen Shot 2019-12-09 at 22 34 19

😄


mfix22:
Thank you @felipefialho 🙏

PR #922, opened 2020-01-04
Fixed first prefix always selected(hightlighted / cursor "not allowed")

aquaductape:
This pull request fixes the issue that when selecting other presets, the first preset remains highlighted with "not allowed" cursor event and cannot be selected.

Blue preset is clicked, but first grey preset is still highlighted and cannot be selected.
Screenshot from 2020-01-04 09-00-09

Preset selection fixed.
Screenshot from 2020-01-04 08-59-24

In presets file, selected is based on the value of property preset from the default settings object. All preset objects do not have this property but have id instead, since default settings has property preset, all the presets inherit the same property preset, which is why the first preset was always highlighted. Fixed by replacing key name preset to id in default settings.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1052,8 +1052,7 @@ export const DEFAULT_SETTINGS = {
   exportSize: '2x',
   watermark: false,
   squaredImage: false,
-  hiddenCharacters: false,
-  id: 'preset:4'
+  hiddenCharacters: false
 }
 
 export const DEFAULT_PRESETS = [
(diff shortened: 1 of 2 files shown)


lib/constants.js
```diff
@@ -1053,7 +1053,7 @@ export const DEFAULT_SETTINGS = {
   watermark: false,
   squaredImage: false,
   hiddenCharacters: false,
-  preset: 'preset:4'
+  id: 'preset:4'
```
mfix22: I believe you want to just remove this key altogether

aquaductape: You're right, I added another commit to remove the key from default :)

PR #925, opened 2020-01-07
Responsive carbon

aquaductape:
Toolbars overflow the carbon editor container before 768px media query. Changed media query to 920px.
Screenshot from 2020-01-07 10-44-51

The width of the main container is larger than the viewport resulting in horizontal scrollbar and carbon editor not centered. Removed min-width so that main container fits with viewport.
Screenshot from 2020-01-07 10-51-48

Main page has unnecessary horizontal scrollbar even though the content is within the viewport visually due to extra large margin-bottom on the body tag. Removed margins on body tag and applied same amount of margin top on main container and lessened margin bottom on main container.
Screenshot from 2020-01-07 11-07-46

In Firefox, the items in the second toolbar do not match it's parent's flex container height. Changed flex property to include auto so that items height matches parent container.
Screenshot from 2020-01-07 09-58-03


mfix22:
Thanks for an amazing PR @aquaductape! What can I do to get more PRs out of people like you! 😍

components/Page.js
```diff
@@ -42,7 +43,6 @@ class Page extends React.Component {
             @media (min-width: 1024px) {
               .main {
                 ${COLUMN};
-                min-width: 1080px; /* temporary fix for mobile overflow issue */
```
mfix22: @aquaductape does this line have to be removed? I can't remember why, but I intentionally included it here: https://github.com/carbon-app/carbon/commit/10cc6fdb935437a9e153c4b275198780644543e3#diff-2465a0749a2041e522b6b8d7681adf5aR31-R37

What do you think?

aquaductape: It was probably there when it depended on a different layout. After removing it, I made sure the layout didn't break throughout the viewport width. I also checked mobile devices viewport on chrome(even though it's not 100% accurate) and on my galaxy s10e and didn't see any breaking layout.

PR #936, opened 2020-01-24
Add @carbonshbot to community section on README.md

rjmunhoz:
Add a link to a Telegram bot which uses puppeteer to interact with the website and generate an image everytime it receives text wrapped in three backticks (```), as well as a gist URL.

Supports options for changing the overall appearance of the genrated image.
Also supports manually specifying a language as the first line of the code block.

This is how it looks:
image


mfix22:
Thanks @rjmunhoz !

PR #949, opened 2020-02-15
add 'width' option for Carbon container

### LINKED ISSUE #880: Add manual config for image width

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I add `width` option for Carbon container width.

When `Auto-adjust width` is false, `width` slider is showed up.
default value of `width` is 680px.

Closes #881
Closes #880


b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I just found out this PR is duplicated with #881.

Please consider one of PRs to be merged.

mfix22:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs I would be happy merge either PR if the width adjustment was handled by dragging the edge of the window, as I mention here: https://github.com/carbon-app/carbon/pull/881#issuecomment-545584836 (cc @mearns)

mearns:
(Re #881) Sorry, I haven't had a chance to try playing with drag handles at all. Tbh, it's counter to what I was hoping Carbon would provide for me, which is a way to produce consistent code snippet images for blog posts and presentations. Too many settings are automatic or imprecise for my use case, leading to a lack of consistency between images. I understand how drag handles would be convenient for more casual use cases, but it doesn't suit my needs, so I probably won't be putting time into it. Hopefully someone else can if this feature is useful to them =)

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 @mearns Thanks for the good comments.

There are three problems with this PR.

One is the preference of the dragging feature mentioned,

The second is that it is inconsistent with the existing setting method when only the dragging is introduced.

The rest is the problem of changing the position of the menu when adjusting the width using the slider mentioned [here](https://github.com/carbon-app/carbon/pull/881#issuecomment-545610002). You can also check this problem in the GIF file above.

What about the following?

If we provide the **input field** to change the width directly instead of the slider, and also provide a dragging feature, we would all be happy. Whichever method user use, `width` must always be synchronized.

If you all agree, I'll implement it in this PR.

mfix22:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs I agree with everything you mentioned, except for the idea of having the width adjustment in _both_ places. I would like the width adjustment to only be possible by dragging the container.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 Thank you. But I can not agree. It should be possible in both ways. The dragging method is convenient but it is difficult to move the mouse precisely to create the desired pixel. And there is a possibility that setting methods may be inconsistent and users may not realize that they can adjust the width through width dragging.

mearns:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs I agree with you, but I feel like such is the case for the entire carbon interface. Even the existing sliders don't give you precise control or even indicate their current values.

It goes back to the point I made above that I don't think this application is designed for the type of use case where precision or consistency is required (e.g., blog posts, presentation, social media campaigns). I'm not sure what exactly that leaves; a lot of one-offs, I guess.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mearns I think a little differently. Whenever you change a setting value, the querystring in the address bar reflects the current value. You can copy this address and use the same style later.

So I think corbon is designed to maintain a consistent style.

If so, showing the precise current value at the setting will be convenient for the user who wants it, but I don't think it will harm the current carbon direction.

mfix22:
For what it is worth, I am not completely sold on the idea of using sliders for _any_ of our current settings, it was just the best option at the time we implemented those settings. 

Since we are implementing a _new_ setting here, I would like us to use the best UI tactic for this job, which IMO is a draggable edge in this case.

mearns:
That's interesting, I do not have that experience with the address bar @b6pzeusbc54tvhw5jgpyw8pwz2x6gs . The only thing that happens in my address bar when I use carbon is that it auto-saves and gives me some kind of unique identifier when I make the first change; the identifier is opaque afaict and doesn't change at all as I edit settings.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
> Whenever you change a setting value, the querystring in the address bar reflects the current value. 



@mearns Please refer to the value of `pv`

@mfix22  I totally agree that dragging is a better UX.
Furthermore, with dragging, if we give the user the option to set the `width` to a precision px by inputting, it would be better than having only one dragging feature. I wonder why you insist on only dragging.

mearns:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs Yup, I understand the concept, but mine doesn't do that. I tried it with and without being signed in and all it gives me if a unique ID for the snippet.

Actually, I tried it incognito and it works as you showed, but I'm not really clear why a) this functionality wouldn't be present for signed in users; or b) why I need to go incognito even after signing out.

Screen Shot 2020-03-02 at 1 19 59 PM

mfix22:
> a) this functionality wouldn't be present for signed in users; or b) why I need to go incognito even after signing out.

@mearns if you are noticing a separate problem, please open a new issue for it.

mfix22:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs if you truly think having the input in the Settings is a _requirement_, then I am open to reevaluating my opinion. 

However, we must also support the drag functionality in order to support this feature. 

To your other point, how would a slider enable a user to set pixel-precise values any more than dragging the edge? Both UI elements should operate in the same fashion: meaning if you click the edge and then use the right and left arrow keys, the width should be adjusted by 1px (same as the slider).

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 Sorry for the late reply. Thank you for thinking positively.

> To your other point, how would a slider enable a user to set pixel-precise values any more than dragging the edge? 

How about displaying the current value along with the slider like the `First line number` value?

Screen Shot 2020-04-06 at 2 01 05 AM

mfix22:
> How about displaying the current value along with the slider like the `First line number` value?

Hmm. What about this: single number input (without slider) in the menu (same component as "First line number"), for direct width entry, and then edge of the container can be used as the "slider" component.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 
I think it's also good, and I agree.
Then I'll re-write the code again with the below two and request a review again!

-Dragging the container edge to adjust the width
-Single number input in the menu

mfix22:
Hi @b6pzeusbc54tvhw5jgpyw8pwz2x6gs 👋 are you still planning to work on this feature? 

No rush if you are, just wanted to check in to make sure I don't prohibit others from contributing to this work (e.g @mearns)

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 I'll push the draft by June 8th. Sorry it took so long.

mfix22:
> @mfix22 I'll push the draft by June 8th. Sorry it took so long.

Don't apologize! There is no rush 🙂Appreciate your help here.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@mfix22 I added a draggable edge to adjust the width. Review the new commit.

mfix22:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs awesome. I won't have a chance to look at this PR this week, but will try and get to it this weekend. Appreciate your efforts thus far.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I added a commit. Now the `widthAdjustment` (auto-adjust width) is automatically set to false when the edge starts dragging.

mfix22:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs this is absolutely incredible. Thank you SO much for all your work here.

PR #965, opened 2020-03-12
fix the broken gitpod setup.

### LINKED ISSUE #963: Gitpod Automated Setup for contributing is broken.
I'll shortly make a Pr with the fix.

image

nisarhassan12:
Fixes #963 

You can test the setup fix on my fork of the repo via this link:

https://gitpod.io/#https://github.com/nisarhassan12/carbon

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,9 @@
+FROM gitpod/workspace-full-vnc:latest
+
+# Pin Node.js to v12.
+
+RUN bash -c ". .nvm/nvm.sh \
+    && nvm install 12 \
+    && nvm use 12 \
+    && nvm alias default 12 \
+    && npm install -g yarn"
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -654,6 +654,16 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "nisarhassan12",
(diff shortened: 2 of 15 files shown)


.gitpod.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM gitpod/workspace-full-vnc:latest
```
mfix22: I don't want to include this file. If there is a simple GitPod setup, I am happy to maintain it in Carbon.

It is my understanding from what @anudeepreddy  said here that all we need to do is update the node version our CI runs on.

anudeepreddy: hey @mfix22 , I think there has been a misunderstanding. I just wanted to know if we could run the development setup on node 10.x as your CI was using the same. 

As you have clearly mentioned that you want all the development to happen in node 12.x, this is the PR that @nisarhassan12 could come up with. If you totally don't want to add the .gitpod.Dockerfile to the setup, It's totally fine. I will try to get in touch with the team at Gitpod and try to find an alternate solution with just the .gitpod.yml file. Thank you for your patience, It's highly appreciated.

mfix22: @anudeepreddy understood. You are correct, I do not want to support a custom Dockerfile for Gitpod — it should work appropriately on Node's current LTS version at the very least. 

I will reopen #963

PR #990, opened 2020-04-16
Update theme variable color definitions

### LINKED ISSUE #987: Theme configurations need updating with variable-2 and variable-3 options
## Syntax highlighting with php seems broken

Whatever the theme I choose, syntax highlighting for PHP seems broken.

This is what it should show (as seen on [codemirror demo page](https://codemirror.net/mode/php/index.html)). It shows two different colors for variables and methods (blue and black):

79136572-86a70e80-7db1-11ea-8307-78760c8044a1

But this is what I get with "dracula" theme on carbon. It only shows one color (white) for both variables and methods:

79136596-90c90d00-7db1-11ea-8fdf-c758f179db19

This is an extract from theme's css, we should get two different colors:

```
.cm-s-dracula span.cm-variable { color: #50fa7b; }
.cm-s-dracula span.cm-variable-2 { color: white; }
```

but while inspecting I get something different:

79137460-0c778980-7db3-11ea-8186-1f537e44e015

Here is a live example, but I noticed it on most php snippets I saw on twitter: https://carbon.now.sh/1HBRbot05AIFiezO6H2n

mfix22:
@smknstd this should fix the highlighting you mentioned for all themes
Closes #987


smknstd:
cool, did you fix it for all themes are do you still need help for some of them ?

mfix22:
@smknstd I went through every theme, and fixed all the ones I noticed were wrong. If any are missing, it is because I made a mistake 😅If you notice one (or more), please feel free to raise a new issue, or fix it directly! 👍 

Also, I have an idea that I will follow up on that won't override the styles unless it has to. More to come…

PR #1012, opened 2020-05-17
chore: move prettier to dev deps

### LINKED ISSUE #1011: Prettier could be moved to the development dependencies
I was looking through the app to learn some Next.js, and wondered why some of the production dependencies were there. 

It looks like _prettier_ is actually not needed in production, but rather only during development.
It allows to shave 10Mb off the production bundle deps, at no cost. 

I ran the integration tests, ran and built the app locally to test everything and saw no behaviour difference. Of course do feel free to correct me if that would break anything 

See #1012

jlengrand:
This very minor PR moves prettier to the dev dependencies, as it is not used in the code anywhere. 

It allows to shave 10Mb off the production bundle deps, at no cost. 

I ran the integration tests, ran and built the app locally to test everything and saw no behaviour difference. Of course do feel free to correct me if that would break anything 

Closes #1011

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "ms": "^2.1.2",
     "next": "^9.3.6",
     "next-offline": "^5.0.2",
-    "prettier": "^2.0.5",
     "react": "^16.13.1",
     "react-click-outside": "^3.0.0",
     "react-codemirror2": "^7.1.0",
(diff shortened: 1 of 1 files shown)


mfix22:
@jlengrand we use `prettier` at runtime: https://github.com/carbon-app/carbon/blob/419070e895f17d7fd8815127c2e1347332f1fc2a/lib/util.js#L83

jlengrand:
Oh, I missed that indeed. Apologies!

mfix22:
@jlengrand no problem! Thanks for closing.

PR #1020, opened 2020-05-21
New Carbon Spinner

albseb511:
Added a carbon spinner, took inspiration from some of my chemistry classes.

spinner

closes: https://github.com/carbon-app/carbon/issues/1004

Any improvements do let me know.

If this approach is good, 

I can make some more edits to accept props and change size accordingly

### DIFF
--- a/components/Spinner.js
+++ b/components/Spinner.js
@@ -1,0 +1,68 @@
+import React from 'react'
+
+const Spinner = () => {
+  return (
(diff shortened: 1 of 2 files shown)


mfix22:
Hey @albseb511, this approach won't work too well when within a button.

Let's discuss options in the issue thread more before opening a PR!

PR #1091, opened 2020-09-02
Keymaps (sublime,vim,...)

Aslemammad:
I added a keymaps option for people who love vim, emacs, or sublime(the ordinary one). Love.

### DIFF
--- a/components/Carbon.js
+++ b/components/Carbon.js
@@ -21,6 +21,7 @@ import {
   LANGUAGE_MIME_HASH,
   DEFAULT_SETTINGS,
   THEMES_HASH,
+  KEY_MAPS,
 } from '../lib/constants'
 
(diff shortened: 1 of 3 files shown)


Aslemammad:
@kavu

mfix22:
Hi @Aslemammad sorry for the delay. 

Carbon is not meant to replicate other editors—it has entirely separate goals than a code editor does—and therefore we want it to behave differently. 

That said, I might be fine with adding a Vim mode toggle in the Editor menu, as a beta feature, to see if other people are interesting in using it, but we won't support a dropdown list of keybindings.

Aslemammad:
I got that, Thanks for preview mode, I appreciate it.

Aslemammad:
We'll see the users' reactions.

Aslemammad:
Or we can add vim as a checkbox.

PR #1097, opened 2020-09-10
Fix encoding % sign breaking SVG background

### LINKED ISSUE #1095: Carbon SVG Export has Broken Backgrounds
**Describe the bug**
The SVG export should look somewhat similar to the PNG exported version. Instead it has a white background, which isn't expected, and is almost unreadable.

image

Instead, it looks like this: 

image

**To Reproduce**

1. Go to https://carbon.now.sh
image

2. Make sure there's some code in the view
3. Click export as SVG

image

This seems to happen with any BG colour, any code, any syntax highlighting theme, including the defaults when you first navigate to the page.

- OS [e.g. macOS, Linux, Windows, iOS]: Reproducible on Linux, Windows and MacOS.
- Browser [e.g. Chrome, Safari, Firefox]: Reproducible on Vivaldi and Firefox.
- Carbon URL [e.g. carbon.now.sh?bg=pink]: https://carbon.now.sh/

Here is a copy of the .svg that is exported. Stored within a .zip file as GitHub issues can't embed SVG files.

## [carbon.svg.zip](https://github.com/carbon-app/carbon/files/5195284/carbon.svg.zip)

<details>
  <summary>Code snippet</summary>
  <pre>
  
   const pluckDeep = key => obj => key.split('.').reduce((accum, key) => accum[key], obj)

const compose = (...fns) => res => fns.reduce((accum, next) => next(accum), res)

const unfold = (f, seed) => {
  const go = (f, seed, acc) => {
    const res = f(seed)
    return res ? go(f, res[1], acc.concat([res[0]])) : acc
  }
  return go(f, seed, [])
}
  </pre>
</details>

qw-in:
This is a quick fix for the backgrounds breaking - guess I'll have to spend more time with this whole encoding issue than initially anticipated. Really hoping to find some time to put together those tests but it's been sunny here :sweat_smile: 

Escaping the `%` sign was breaking some of the styles, specifically percentage units. But only for SVG :cry: 

Closes #1095

### DIFF
--- a/lib/dom-to-image.js
+++ b/lib/dom-to-image.js
@@ -65,7 +65,8 @@
         return makeSvgDataUri(
           clone,
           options.width || util.width(node),
-          options.height || util.height(node)
+          options.height || util.height(node),
+          options.escapePercentSign
         )
       })
 
@@ -101,7 +102,9 @@
    * @return {Promise} - A promise that is fulfilled with a PNG image data URL
    * */
   function toPng(node, options) {
-    return draw(node, options || {}).then(function (canvas) {
+    options = options || {}
+    options.escapePercentSign = true
+    return draw(node, options).then(function (canvas) {
       return canvas.toDataURL()
     })
   }
@@ -124,7 +127,9 @@
    * @return {Promise} - A promise that is fulfilled with a PNG image blob
    * */
   function toBlob(node, options) {
-    return draw(node, options || {}).then(util.canvasToBlob)
+    options = options || {}
+    options.escapePercentSign = true
+    return draw(node, options).then(util.canvasToBlob)
   }
 
   function copyOptions(options) {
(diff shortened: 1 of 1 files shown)


lib/dom-to-image.js
```diff
@@ -545,8 +552,11 @@
       return array
     }
 
-    function escapeXhtml(string) {
-      return string.replace(/%/g, '%25').replace(/#/g, '%23').replace(/\n/g, '%0A')
+    function escapeXhtml(string, escapePercentSign) {
+      if (escapePercentSign) {
```
mfix22, line 556: @qw-in I'm pretty sure that instead of passing this in as a prop, we can just update the code path that converts to PNG to escape the percent sign, and don't convert that in the path used by SVGs.

mfix22 (changes requested):
@qw-in really appreciate you looking in to this. Would love to look into a solution that avoids the extra option.

mfix22:
@qw-in merged this to get the fix out the door, but feel free to come back to simplify it. I would prefer to avoid boolean options like `escapePercent` if possible. But if it is not, no worries.

Really appreciate your continued help here! It is a huge help to Carbon 🙂

PR #1107, opened 2020-10-01
Fix typo in scrollbarStyle option

boyvanamstel:
The option would be ignored when exporting the project through `next build && next export`, failing to hide the vertical scroller.

(Additional changes were presumable made automatically by the linter.)

Screen Shot 2020-10-01 at 11 04 56


mfix22 (approved):
Thank you!

PR #1114, opened 2020-10-03
Dutch translation of README

timwienk:
README in Dutch! :netherlands:

I wasn't entirely sure where to add the Dutch document to the Translations table in the English README.md. Let me know if you're happy with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,9 @@
       <td>
         <a href="/docs/README.tr.md">Türkçe</a>
       </td>
+      <td>
+        <a href="https://github.com/carbon-app/carbon/new/main/docs?filename=docs/README.%3Cnew%20language%3E.md">Add +</a>
+      </td>
     </tr>
     <tr>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -57,6 +57,9 @@
       <td>
         <a href="/docs/README.tr.md">Türkçe</a>
       </td>
+      <td>
```
mfix22, line 60: Can you move this to be next to the "Add +" `td`? 🙂 

There is no real method to the ordering, besides just keeping them the order in which they have been translated

mfix22 (commented):
This is great! Just one nit

Thanks @timwienk

README.md
timwienk: Like this? (Swapped "Add +" and "Nederlands" around.)

mfix22: Works for me!

mfix22 (approved):

PR #1131, opened 2020-11-16
added Taiwanese version README

JeremyShih:
Added a new README version in Taiwanese language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,10 @@
         <a href="/docs/README.tr.md">Türkçe</a>
       </td>
       <td>
-        <a href="/docs/README.tw.md">臺灣</a>
+        <a href="/docs/README.kr.md">한국어</a>
       </td>
     </tr>
     <tr>
-      <td>
-        <a href="/docs/README.kr.md">한국어</a>
-      </td>
       <td>
         <a href="/docs/README.ta.md">தமிழ்</a>
       </td>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -84,6 +84,7 @@
         <a href="/docs/README.nl.md">Nederlands</a>
       </td>
       <td>
+        <a href="https://github.com/carbon-app/carbon/new/main/docs?filename=docs/README.%3Cnew%20language%3E.md">Add +</a>
```
mfix22, line 87: Can you actually shuffle these around such that the Taiwanese version comes just to the left of the "Add +" one (instead of above it)?

docs/README.tw.md
```diff
@@ -0,0 +1,341 @@
...
+  <a href="https://opencollective.com/carbon-app#sponsors">	
+    <img src="https://opencollective.com/carbon-app/tiers/sponsors/badge.svg?label=Sponsors&color=297eff" alt="Open Collective sponsors" />	
+  </a>
+</p>
+
+<br></br>
+
+##### 翻譯
+
+<table>
```
mfix22, line 36: Please remove the translation table here 🙂

mfix22 (changes requested):
This is looking fantastic (AFAICT) @JeremyShih! I just left a couple tweaks, and then this will be good to go. 

Thanks a lot for doing this. Once this gets merged, I'll add you as a `translator` contributor 🙂

README.md
JeremyShih: image
to put it this way?

docs/README.tw.md
JeremyShih: removed

README.md
mfix22: Yup! That's perfect

mfix22 (changes requested):
One last thing: can you add `docs/README.tw.md` to the `.all-contributorsrc`?

Then this will be good to go!

JeremyShih:
Added path to the .all-contributorsrc

mfix22 (approved):

PR #1168, opened 2021-02-04
Add support for Copy As HTML

### LINKED ISSUE #1042: Copying to clipboard
Is there anyway to make it so you can export it as some equivalent of RTF so I can paste the code in with the colour coding? At the moment I use SublimeHighlight in Sublime Text to do this as it has a 'copy to clipboard as RTF'. If that were possible on Carbon it would be great.

Arriv9l:
I think **Copy As HTML** might has the same effect as **Copy As RTF**.
[copy-to-clipboard](https://github.com/sudodoki/copy-to-clipboard#api) now support copy with options, so I’ve already submitted a  [PR](https://github.com/mfix22/actionsack/pull/10) to [actionsack](https://github.com/mfix22/actionsack).
The effect is as follows, I’ve tested it by modifying node_modules\actionsack\dist\index.esm.js.

Close #1042

### DIFF
--- a/components/CopyMenu.js
+++ b/components/CopyMenu.js
@@ -23,6 +23,79 @@ const toURL = url => `${location.origin}${url}`
 const replaceAsterisks = string => string.replace(/\*/g, '%2A')
 const toEncodedURL = morph.compose(encodeURI, replaceAsterisks, toURL)
 
+function getTokenColors(config) {
+  const colors = {}
+  for (let token of document.querySelectorAll(`[class^=cm-]`)) {
+    let className = token.className
+    if (!colors[className]) {
+      colors[className] = getComputedStyle(token).color
+    }
+  }
+  if (config.lineNumbers) {
+    let cmLineNumber = document.querySelector(`.CodeMirror-linenumber`)
+    colors['cm-linenumber'] = getComputedStyle(cmLineNumber).color
+  }
+  return colors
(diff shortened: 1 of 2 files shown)


mfix22 (commented):
@Arriv9l this is super cool! I'm happy to adopt this, but can we make sure the entire image gets copied? Not just what is within the text box? We want to include the background as well, for example.

Arriv9l:
@mfix22  Sure. I'll give it a try.

Arriv9l:
The final result is as follows:


- [x] Window Control: SVG is not supported, so I switched to emoji 🟠🟡🟢.
- [x] Background Color: has a limit of up to 6pt
- [ ] Background Image: not supported
- [ ] Drop Shadow: not supported

mfix22:
@Arriv9l sorry if I wasn't clear in my previous statement. If we are to support this feature, it must match the live snippet exactly: it can't merely mimic it 😞 Sorry about that. 

I'm going to close this PR for now, but if you find a solution that allows for copying the HTML as is (styles and all) I will happily support it!

Arriv9l:
Got it! 
For now, I’ve written a [script](https://greasyfork.org/en/scripts/422037-carbon-copy-as-html) to fulfill my requirements. 
If anyone has a similar need, feel free to try.

Issue #1169, opened 2021-02-05
Replace axios with redaxios


mfix22:
https://github.com/developit/redaxios

PR #1170, opened 2021-02-07
Switches Axios to Redaxios

mlucap:
## Fixes

Fixes and closes issue #1169 

## Description

Switches Axios to Redaxios.
This seems like the correct way to go about changing as I was viewing commits on other repositories doing it the same way.


mfix22:
Thanks!

PR #1173, opened 2021-02-21
Fix broken "Sponsor" button

ajs256:
- [N/A] Integration tests (if applicable)

Closes # N/A

Still having a `.github/FUNDING.yml` shows a "Sponsor" button without any links. This removes the funding file, to reduce confusion.


ajs256:
Looks like the CI fail was Vercel complaining that "authorization is required to deploy". Should I worry about that?

mfix22:
Great catch! We were in the middle of enabling GitHub Sponsors, and hadn't updated the button now. 
It has since been fixed!

PR #1192, opened 2021-03-07
Don't sanitize payload for storage

### LINKED ISSUE #1183: Auto-adjust width is apparently broken
**Describe the bug**
Once you disable **auto-adjust width**, It never goes back again. If you try to enable again, it seems like it's worked, however if you update the page, auto-adjust is turned off again.

**To Reproduce**

1. Login with github
2. Create code snippet
3. Disable auto-adjust width
4. Enable auto-adjust width again
5. Reload page or open link in another tab

**Expected behavior**
It should keep the configuration (should keep auto-adjust width turned on)

**Screenshots**
When editing on Carbon page (looks fine):
[
Screenshot 2021-03-04 at 13 35 15
](url)

- OS [e.g. macOS, Linux, Windows, iOS]: MacOS
- Browser [e.g. Chrome, Safari, Firefox]: Opera, Chrome

adghayes:
Reverts the changes of commit 520e19ab44065622a7a47c22d60ccd70de0e68bc. Setting changes to default values are no longer removed from the Firebase payload, avoiding the bug described in #1183. Passes E2E tests but I don't have a means to test Firebase network requests locally - @mfix22 can you share a deployment link, or test yourself if that's easier.

Closes #1183, although another change is required on backend to avoid storing default values.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -1,11 +1,10 @@
 import axios from 'axios'
 import debounce from 'lodash.debounce'
-import omitBy from 'lodash.omitby'
 import ms from 'ms'
 
 import { fileToDataURL } from './util'
 import firebase from './client'
-import { DEFAULT_CODE, DEFAULT_SETTINGS } from './constants'
+import { DEFAULT_CODE } from './constants'
 
 export const client = axios.create({
   baseURL: `${process.env.NEXT_PUBLIC_API_URL || ''}/api`,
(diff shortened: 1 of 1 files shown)


mfix22:
@adghayes I moved the sanitization to the backend, which uses [`admin.firestore.FieldValue.delete()`](https://firebase.google.com/docs/firestore/manage-data/delete-data) to remove all fields that are the defaults

If you want to test it out, here is the staging deployment: https://carbon-mfix22-carbon-app.vercel.app

adghayes:
I can't seem to save a snippet with this deployment, but I confirmed that the outbound POST requests now contain fields that need to be reset to defaults. It's a significant but I think fine amount of additional data over the wire, and it should fix our problem.

mfix22:
> I can't seem to save a snippet with this deployment

Ahh yes, there is a bug in my deployment. Will fix that before merging here.

mfix22:
Link above should be working now 🤞

adghayes:
thanks @mfix22. with the deployment fixed, I was able to confirm that the bug described in #1183 is fixed by this patch. As long as you're OK with the larger payloads (which I think are fine), this is looking good

PR #1193, opened 2021-03-08
Added contributors page 🎉

heybereket:
Hey Max!

You'll see me contrbuting here way more often but anyway - I've gone ahead and added a contributors page to the site, it basically fetches the Github API and returns all the contributors for this repository and displays them!

The page the change is on is: /contributors 
I also went ahead and added the page to the footer as well! ✨

Here's a screenshot:


components/Footer.js
```diff
@@ -19,6 +19,9 @@ const Footer = () => (
       <a className="link" href="https://github.com/carbon-app/carbon">
         source
       </a>
+      <a className="link" href="/contributors">
```
mfix22, line 22: I'm not sure I want to add it to the footer yet. Can we add it to the "About" page?

pages/contributors.js
```diff
@@ -0,0 +1,113 @@
...
+  
+        setContributors(
+          contributors.filter((contributor) => !contributor.login.includes('bot'))
+        )
+      }
+  
+      fetchContributors()
+    }, [])
+  return (
+    <Page>
```
mfix22, line 22: This page looks great 🙂

pages/contributors.js
```diff
@@ -0,0 +1,113 @@
+import { React, useState, useEffect} from 'react'
+import Page from '../components/Page'
+
+export default function About() {
+    const [contributors, setContributors] = useState([])
+    
+    useEffect(() => {
+      const fetchContributors = async () => {
+        const response = await fetch(
+          'https://api.github.com/repos/carbon-app/carbon/contributors'
```
mfix22, line 10: I think you will want to increase the page size here, because right now it is only pulling the first 30

```
https://api.github.com/repos/carbon-app/carbon/contributors?per_page=100
```

We actually have over 100 contributors, though.

pages/contributors.js
```diff
@@ -0,0 +1,113 @@
...
+        )
+        const contributors = await response.json()
+  
+        setContributors(
+          contributors.filter((contributor) => !contributor.login.includes('bot'))
```
mfix22, line 15: I think you will want to filter by logins ending in `[bot]` — right now this will even filter legitimate users, like someone names `bottom`, for example 🙂

mfix22 (changes requested):
This is pretty neat! Left a couple pieces of feedback

pages/contributors.js
heybereket: Makes sense!

pages/contributors.js
heybereket: Thank you!

components/Footer.js
heybereket: Yeah for sure, let me add it under the about page!

pages/contributors.js
heybereket: Ahh got it! I don't think Github has set a limit for the "per page", so for now I just set it as 500, I will look into it more and see if there is a max/full option :)

heybereket:
> This is pretty neat! Left a couple pieces of feedback

Thank you! Just committed the updated changes - should be all good now! 
- Added the contributors page to the about page
- Made all the contributors show 
- Made a few tweaks to the text on the about page where the contributions section is (at the bottom)

mfix22 (approved):

PR #1195, opened 2021-03-08
Share to Twitter as text [WIP on screenshot] 🚧✨

heybereket:
Added in a share option where it would open up a new tab where someone can tweet about carbon - it looks like the images below:
However, I'm still working on the "Tweet" button so that a screenshot of the code snippet on the editor can be attached. 

image
image


heybereket:
Woops! Forgot to switch to different branches haha, my bad! Will remember to do that for next time ✨

mfix22:
Hey @heybereket — this is something we don't what do support at the moment. We want to keep the main application controls focused on utility of generating screenshots. Sorry for the extra work 😥

Issue #447, opened 2018-08-03
New default code snippet


mfix22:
Hey Carbon users,
We want a new default code snippet (or maybe a few) and need your help. 

What should we make it? 🙂Comment below with suggestions.

mittalyashu:
What do you mean by default code snippet?

mittalyashu:
I got one.

```js
var iAmAwesome = true;

if iAmAwesome {
   open(carbonNow)
} else {
  open(localScreenshotTool)
}
```

mfix22:
```jsx
// Built with Carbon
<Dropper
  onDrop={files => this.setState({ files })}
  filter={file => file.size < 10000 && file.type === 'image/png')}
>
  {({ isOver, canDrop, files }) => (
    <Overlay
      title={`Drop your files here to import ${isOver ? '✋' : '✊'}`}
    >
      <Upload files={files.map(file => file.content)} />
    </Overlay>
  )}
</Dropper>
```

mfix22:
@MONIKA-KUMAR-JETHANI feel free to make suggestions in the comments of this issue! 🙂

sdabhi23:
Should the same snippet be used for all languages or should the snippet change when the language is changed from the menu? 🤔

mfix22:
I think it would overcomplicate things to switch the code when the language changes since we would have to keep track of if the user has typed anything before switching. I would love to replace the current one with a fresh JavaScript/React one, since that is mostly what Carbon was build with 🙂

v-fernandez:
How about something like:

```javascript
const ThemeContext = createContext({ theme: "light", toggleTheme: () => {} })

const ThemeProvider = props => {
  const [theme, setTheme] = useState("light")
  const toggleTheme = () => setTheme(theme === "light" ? "dark" : "light")

  return (
    <ThemeContext.Provider value={{ theme, toggleTheme }}>
      {props.children}
    </ThemeContext.Provider>
  )
}
```

mfix22:
@v-fernandez I like this^ ! I think we should include something like this in our next array of default snippets

PR #1196, opened 2021-03-08
Random Code Generation 🚀👀

heybereket:
Feel free to edit the random code that gets generated by default by I've seen a lot of code snippets like Carbon have random code snippets generate onload (for ex. ray.so) - I added that to Carbon :D 

I put all the code snippets inside of an array: snippets_array, and it would generate a random one from there and display it - feel free to change it if you feel the need/leave any feedback for any improvements you feel should be needed/added! ✨

Example of a generated code snippet by random: 👇🏻

image


heybereket:
This time I made a branch for this feature :)
I'm also looking more at the issues and trying to solve some of them :D

heybereket:
Note: the padding/margins on the image might look weird - not sure what's going on there but when I try it on the main site: carbon.now.sh, it seems to work fine with the styling all looking good!

mfix22:
Hey @heybereket, I want to be more careful with which snippets we show if we were to to update the default one. I would want them to mostly take the same shape as the current one, and that is difficult to do without careful consideration.

If you have other ideas for them, feel free to leave a comment in #447

PR #1199, opened 2021-03-25
Hebrew translation of README.md

Guy-Adler:
Added a new README version in Hebrew


mfix22 (changes requested):
Hey @Guy-Adler! Thanks for making this translation 🙂 

Could you add a link to this translation in the translations table here: https://github.com/carbon-app/carbon/blame/main/README.md#L81

And also add the README name to the all-contributorsrc list list here: https://github.com/carbon-app/carbon/blob/main/.all-contributorsrc#L22

After that, I would be happy to merge!

(Also heads up, you might end up with a merge conflict if I merge https://github.com/carbon-app/carbon/pull/1200 first)

mfix22 (approved):

PR #1200, opened 2021-03-27
Add Italian README.it.md

LorenzoLancia:
Added Italian translation for README.md

I created a translation of the Readme in Italian because I saw that it was missing

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -9,6 +9,7 @@
     "docs/README.es.md",
     "docs/README.fr.md",
     "docs/README.hi.md",
+    "docs/README.it.md",
     "docs/README.ja.md",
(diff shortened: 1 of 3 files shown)


mfix22 (changes requested):
Hey @LorenzoLancia! Thanks for making this translation 🙂 

Could you add a link to this translation in the translations table here: https://github.com/carbon-app/carbon/blame/main/README.md#L81

And also add the README name to the all-contributorsrc list list here: https://github.com/carbon-app/carbon/blob/main/.all-contributorsrc#L22

After that, I would be happy to merge!

mfix22:
(Also heads up, you might end up with a merge conflict if I merge #1199 first)

mfix22 (approved):

PR #1223, opened 2021-05-27
Update ExportMenu.js

neerajcodes:
I want to add some functionality then anybody download image then they wiill be not facing 
Carbon.png already exists 
Because i use Random string Function to generate random strings as your wish 8 Bit 16 Bit 4 Bit 
Because i like carbon.now.sh application

Closes #


mfix22:
Hey @neerajcodes. Sorry for the delay here. We actually prefer to keep the images names constant, since you can choose to change the name in the file dialog, and operating systems like MacOS will automatically append a number like `carbon (1).png`

Thanks for the suggestion.

PR #1245, opened 2021-08-19
support run carbon app localy by using docker

Zheaoli:
1. support run carbon app localy by using docker

- [ ] Integration tests (if applicable)

Closes #

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+**/.git
+**/node_modules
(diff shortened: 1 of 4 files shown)


mfix22:
Hey @Zheaoli — thanks for pushing a PR instead of an issue. However, we don't want to support this used case for local dev. Since it is just a simple node app, it's super easy to just create your own Dockerfile (as you have done) or even publish an image to Docker hub.

PR #1250, opened 2021-08-23
Add Cascadia Code font

cyntler:
Adding a new Cascade Code font: https://github.com/microsoft/cascadia-code.


mfix22:
Hey @cyntler — unfortunately, Carbon is no longer accepting new core fonts. If you want to use Cascadia code, please upload a custom font when using the app.

PR #1252, opened 2021-09-07
Upgrade to Next.js 11

jeremynguyencs:
Upgraded to Next.js 11. Used IgnorePlugin() to prevent global css files (which are no longer supported in Next.js 11) from benig generated in webpack.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -870,6 +870,15 @@
       "contributions": [
         "translation"
       ]
+    },
+    {
+      "login": "jeremynguyencs",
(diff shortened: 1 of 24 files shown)


mfix22 (changes requested):
@jeremynguyencs there is still a strange issue where the initial render of the codemirror editor is not displayed correctly. 

image

Any re-render will cause this rendering glitch to go away. 

if you fixed that issue, that would be absolutely outstanding!

mfix22:
@jeremynguyencs if you could reach out to someone from the Next.js community to resolve this, that would be amazing 🙏

PR #1261, opened 2021-10-06
fix no longer valid gist-based carbon url example in `about` page.

rizdaprasetya:
- [x] Prettier lint-ed

The main changes is in this commit: https://github.com/carbon-app/carbon/commit/64cef8b37c41c60489196387deeee2032c46a231 . 
- It fixes: I found on [about page](https://carbon.now.sh/about) that the gist-based url example that seems to be no longer valid url (the original gist get deleted maybe?), which result in the url redirecting to home. So I replace it with the example gist url found in the readme.md
- Also added example of the original gist url VS the result carbon url, to make it easy to understand which ID from the url to copy, and not to include username from the url.

Also in spirit of **Hacktoberfest**, **if you think this PR is valid** and accepted, please kindly add this label to this PR: `hacktoberfest-accepted`. Thank you 🙂 

Btw Carbon rocks! Useful tool indeed!

## Preview of Changes
Screen Shot 2021-10-06 at 19 07 58

### DIFF
--- a/pages/about.js
+++ b/pages/about.js
@@ -76,16 +76,9 @@ export default function About() {
           <ul className="mt0 mb3">
             <li>Drop a file into the editor</li>
             <li>
-              Append a GitHub gist id to the url (e.g.{' '}
-              <a
-                className="link"
-                href="https://gist.github.com/mfix22/3208813b324d82a9ebd197e4b1c3bae8"
-              >
-                original gist url
-              </a>{' '}
-              ➔{' '}
+              Append a GitHub gist id to the url (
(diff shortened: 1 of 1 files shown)


mfix22:
> Cool 👍 Also if it's not too much to ask, would you please add `hacktoberfest-accepted` label on this PR @mfix22 ? thanks!
> 
> > in spirit of Hacktoberfest, if you think this PR is valid and accepted, please kindly add this label to this PR: `hacktoberfest-accepted`. Thank you 🙂

Of course!

PR #1277, opened 2021-11-09
Try optional catch all routing

mfix22:

Closes https://github.com/vercel/next.js/discussions/29856


mfix22:
Doesn't work in this version of Next

PR #1278, opened 2021-11-10
Add Cascadia font

### LINKED ISSUE #1276: Cascadia Code
There's a new open-source font that is now the default in Windows Terminal and Visual Studio: https://github.com/microsoft/cascadia-code

Would you be open to adding Cascadia Code and Cascadia Mono (the non-ligature version which I prefer) to the font list?

I'd be happy to try to figure out how to contribute the change.

jnm2:
Closes #1276, incorporates @cyntler's contribution at https://github.com/carbon-app/carbon/pull/1250

After selecting Cascadia Code:
image

~~After selecting Cascadia Mono (the non-ligature option):~~ **Now removed**
image


lib/constants.js
```diff
@@ -2,6 +2,8 @@ import toHash from 'tohash'
 
 export const FONTS = [
   { id: 'Anonymous Pro', name: 'Anonymous Pro' },
+  { id: 'Cascadia Code', name: 'Cascadia Code' },
```
mfix22, line 5: Please only add one or the other. Which of these versions is the most popular/biggest default? Let's go with that one.

mfix22 (changes requested):

lib/constants.js
cyntler: @mfix22 @jnm2 `Cascadia Mono isn’t really a different font from Cascadia Code. It’s a version of Cascadia Code without the programming ligatures.` - https://www.ctrl.blog/entry/cascadia-system-font.html.

IMO we should to add `Mono`.

jnm2: I don't like ligatures and I worry that it could confuse people when sharing code, so I prefer Cascadia Mono. But honestly I'm not sure I can say it's more popular or more default than Cascadia Code.

And yes, it's one font. Just a way to control ligatures.

cyntler: > I don't like ligatures and I worry that it could confuse people when sharing code, so I prefer Cascadia Mono. But honestly I'm not sure I can say it's more popular or more default than Cascadia Code. Some people love ligatures, and that's e.g. Fira Code's whole thing.
> 
> And yes, it's one font. Just a way to control ligatures.

In our case, it probably doesn't matter which one will be. By default, ligatures will not be supported. So if the font is better known by the name `Cascadia Code` then this choice is ok too.

jnm2: >  By default, ligatures will not be supported.

I'm not sure what this means, since I can see ligatures in the screenshot at the top when I selected Cascadia Code.

cyntler: > > By default, ligatures will not be supported.
> 
> I'm not sure what this means, since I can see ligatures in the screenshot at the top when I selected Cascadia Code.

Oh sorry, my mistake.

jnm2: Much as I prefer Mono, I think the correct answer to @mfix22's popular+default criteria is Cascadia Code.

jnm2:
Dropped the commit adding Cascadia Mono.

cyntler (approved):
👍

PR #1280, opened 2021-11-12
Consolidated type and format into one param

### LINKED ISSUE #1274: Fix this TODO
https://github.com/carbon-app/carbon/blob/6fd6d99a8004f7ce258e27fa26956a526956a5cb/components/Editor.js#L140

Combine `type` and `format` to just use one param: `type`

OpeAbidemi:
Removed the type parameter from the `getCarbon()`  function

Closes #1274

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -96,7 +96,6 @@ class Editor extends React.Component {
   getCarbonImage = async (
     {
       format,
-      type,
       squared = this.state.squaredImage,
       exportSize = (EXPORT_SIZES_HASH[this.state.exportSize] || DEFAULT_EXPORT_SIZE).value,
     } = { format: 'png' }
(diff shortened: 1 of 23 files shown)


components/Editor.js
```diff
@@ -179,7 +178,7 @@ class Editor extends React.Component {
 
     const prefix = options.filename || this.state.name || 'carbon'
 
-    return this.getCarbonImage({ format, type: 'blob' })
+    return this.getCarbonImage({ format })
```
mfix22, line 181: This call is no longer correct. This will now use the `png` path instead of the `blob` path. 

I would update `format = 'blob'` above, and also update where `exportImage` is called in `ExportMenu`

mfix22 (changes requested):
Couple fixes to make @OpeAbidemi. Thanks for working on this.

components/Editor.js
```diff
@@ -200,7 +199,7 @@ class Editor extends React.Component {
   }
 
   copyImage = () =>
-    this.getCarbonImage({ format: 'png', type: 'blob' })
+    this.getCarbonImage({ format: 'png' })
```
mfix22, line 202: Now this one is incorrect.

mfix22 (changes requested):

PR #1313, opened 2022-01-17
fix solidity highlight syntaxing

rubydusa:
the npm package that was used for solidity syntax highlighting didn't run
the code that adds solidity syntax highlight support on import.
The change is a copy of the code by @alincode from the package and now solidity syntax highlighting works


alincode:
@rubydusa maybe you could PR in my repo, so you don't need to do copy here🤔.

mfix22:
> @rubydusa maybe you could PR in my repo, so you don't need to do copy here🤔.

@rubydusa please submit a PR upstream. We do not want to copy @alincode's code.

PR #1322, opened 2022-01-25
eslint config for nextjs

### LINKED ISSUE #1321: Use Next.js recommended lint config
See https://nextjs.org/docs/basic-features/eslint#migrating-existing-config

- [x] Ensure `yarn lint` still passes
- [x] Ensure `plugin:jsx-a11y/recommended` still works
- [x] Ensure `import` ESlint plugin still works
- [x] Ensure config is simpler with the Next.js lint

aymanesarrar:
The next configuration already handles setting default values for the parser, plugins and settings properties. 
more here : https://nextjs.org/docs/basic-features/eslint#migrating-existing-config

Closes #1321


mfix22:
@aymanesarrar should we update the script to be `next lint` instead of `eslint .`? Are we then able to remove a few dependencies?

.eslintrc.js
```diff
@@ -1,19 +1,11 @@
 module.exports = {
-  parser: '@babel/eslint-parser',
-  parserOptions: {
-    requireConfigFile: false,
-    babelOptions: {
-      presets: ['next/babel'],
```
mfix22, line 6: Are we able to remove the various babel/eslint dependencies from our `package.json`?

mfix22:
@aymanesarrar `yarn lint` no longer passes, which is a major requirement to fix issue #1321, as specified in the issue body.

.eslintrc.js
aymanesarrar: no i don't think we should remove it from our package.json

aymanesarrar:
i updated it to next lint and I had a problem that it doesn't know the extends next so I updated it to next/core-web-vitals but It requires to install eslint-next-config I've done it and npm run lint passes with some eslint errors, should i commit ?

PR #1333, opened 2022-02-07
Add separate language mode for F#

baronfel:
CodeMirror's `mllike` mode has sub-support for F# if provided the `text/x-fsharp` MIME type, so I've done so here.

I found this at CodeMirror's documentation for the `mllike` mode [here](https://codemirror.net/mode/mllike/index.html).

Before:
FK9W61oXoAAX0rm

After:
after


baronfel:
Thank you!

Issue #1336, opened 2022-02-09
Question: Do I have to pay for embeds?


Potherca:
In [my blog post on Medium](https://blog.devgenius.io/the-search-for-the-holy-grail-of-code-blocks-on-medium-502071931cac) I speak favorably of Carbon. 

However, one of my readers pointed out to me that embedding Carbon snippets does not seem to work unless one has a paid account. 

This seems to be reflected on the Account page:

![image](https://user-images.githubusercontent.com/195757/153160921-86bb2fa9-d6bf-41dd-9984-908eaac0b5d0.png)

However, I've never had any issues with embedding snippets from my free account. I also noticed that the git commit for the "Carbon Diamond is not quite ready yet." message is 3 years old.

So now I am confused... Is a paid account now required for embedding snippets?

mfix22:
Thanks for the question. That feature is now included in free tiers. Feel free to submit a PR to fix this 🙂

PR #1344, opened 2022-02-20
Change account.js to reflect embeds are free

Potherca:
As stated in #1336, embedding saved snippets is part of the Free tier, but on the account page it is listed as part of the Diamand tier.

This MR fixes that.

### DIFF
--- a/pages/account.js
+++ b/pages/account.js
@@ -80,7 +80,7 @@ function Plan({ selectBilling }) {
           </tr>
           <tr>
             <td>Embed saved snippets</td>
-            <td></td>
+            <td>✔</td>
             <td>✔</td>
           </tr>
           <tr>


pages/account.js
```diff
@@ -80,8 +80,8 @@ function Plan({ selectBilling }) {
           </tr>
           <tr>
             <td>Embed saved snippets</td>
-            <td></td>
             <td>✔</td>
+            <td></td>
```
mfix22, line 84: It needs to be checked in both the free and paid versions:

image

mfix22 (changes requested):
PTAL @Potherca

pages/account.js
Potherca: Doh. :facepalm:

PR #1349, opened 2022-03-13
Fix an issue where exported images would be smaller

### LINKED ISSUE #1287: Web app is unusable on mobile
**Describe the bug**
The mobile version of Carbon is unusable in a few ways. Primarily because there is no mobile version. Scrolling horizontally is unavailable, cutting off many buttons.

**Expected behavior**
A mobile version of the app so one can quickly paste code into and export. 

**Screenshots**
image

- OS [e.g. macOS, Linux, Windows, iOS]: iOS 15.0
- Browser [e.g. Chrome, Safari, Firefox]: Safari and Chrome

hatsu38:

Closes #1287

### Issue Summary
When viewing on mobile, increasing width makes it responsive.
However, the exported image was also responsive and vertically long.
Detail: https://github.com/carbon-app/carbon/issues/1287#issuecomment-1055586425

### Solution
- When the set width is larger than the displayed width, the width of the exported image is calculated using the set width.
- Changed the display from responsive to scrolling
  - https://github.com/carbon-app/carbon/issues/1287#issuecomment-1032739971 
  - > You can no longer view your code, as it is horizontally unscrollable.

https://user-images.githubusercontent.com/16137809/158049188-c1a047de-5080-45a1-8b5d-1c55aa4a0725.mov

### DIFF
--- a/components/Editor.js
+++ b/components/Editor.js
@@ -93,6 +93,16 @@ class Editor extends React.Component {
   updateCode = code => this.updateState({ code })
   updateWidth = width => this.setState({ widthAdjustment: false, width })
 
+  calculateExportImageBaseWidth = offsetWidth => {
+    if (this.state.widthAdjustment) return offsetWidth
+
+    const settingWidth = parseInt(this.state.width)
+    const isOverWidth = offsetWidth < settingWidth
+    if (!isOverWidth) return offsetWidth
+
+    return settingWidth + parseInt(this.state.paddingHorizontal)
+  }
+
   getCarbonImage = async (
     {
       format,
@@ -101,9 +111,9 @@ class Editor extends React.Component {
     } = { format: 'png' }
   ) => {
     const node = this.carbonNode.current
-
-    const width = node.offsetWidth * exportSize
-    const height = squared ? node.offsetWidth * exportSize : node.offsetHeight * exportSize
+    const imageBaseWidth = this.calculateExportImageBaseWidth(node.offsetWidth)
+    const width = imageBaseWidth * exportSize
+    const height = squared ? imageBaseWidth * exportSize : node.offsetHeight * exportSize
 
     const config = {
       style: {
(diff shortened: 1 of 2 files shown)


components/Editor.js
```diff
@@ -93,6 +93,16 @@ class Editor extends React.Component {
...
   updateWidth = width => this.setState({ widthAdjustment: false, width })
 
+  calculateExportImageBaseWidth = offsetWidth => {
+    if (this.state.widthAdjustment) return offsetWidth
+
+    const settingWidth = parseInt(this.state.width)
+    const isOverWidth = offsetWidth < settingWidth
+    if (!isOverWidth) return offsetWidth
+
+    return settingWidth + parseInt(this.state.paddingHorizontal)
```
hatsu38, line 103: When the set width is larger than the displayed width, the width of the exported image is calculated using the set width.

components/Editor.js
```diff
@@ -93,6 +93,16 @@ class Editor extends React.Component {
   updateCode = code => this.updateState({ code })
   updateWidth = width => this.setState({ widthAdjustment: false, width })
 
+  calculateExportImageBaseWidth = offsetWidth => {
```
mfix22, line 96: @hatsu38 sorry for the delay. Unfortunately, I don't want to do any complex calculation here. 

I would prefer for the editor window to go off screen on smaller displays and have this function just capture what is there than try and make the editor fit on smaller screens and do this calculation after the fact. 

Can we just make the editor go off screen on mobile, but allow the user to easily scroll horizontally?

mfix22 (changes requested):
Sorry for the delay @hatsu38

components/Editor.js
hatsu38: @mfix22 Thank you for your comment.
I will fix it at a later date and push it.

mfix22: @hatsu38 no rush on this at all, but just wanted to check in here to see if you have any updates 🙂

hatsu38: Sorry, I have not been able to solve this problem recently due to lack of time

mfix22: No worries at all @hatsu38! I'm going to close this in the meantime, but feel free to re-open it whenever, or open a new PR anytime.

PR #1398, opened 2022-07-23
Update to react@18

OpeAbidemi:
- Upgraded to **react@18**
- Converted **Carbon.js** to function component to use **react@18** hooks
- Used **useDeferedValue** instead of debouncing in **Carbon.js**

@mfix22


components/Carb.js
```diff
@@ -0,0 +1,487 @@
+import React, { useMemo, useState, useDeferredValue } from 'react'
```
mfix22, line 1: @OpeAbidemi what is this extra file being used for?

components/Carbon.js
```diff
@@ -49,14 +46,16 @@ function getUnderline(underline) {
   return 'initial'
 }
 
-class Carbon extends React.PureComponent {
-  static defaultProps = {
+const Carbon = (
```
mfix22, line 49: Why does this component need to get converted to a functional component? If it isn't strictly required for upgrading to React@18, can we do these changes in a separate PR?

mfix22 (changes requested):
Thanks @OpeAbidemi! Just a few questions here and above

Did you confirm that automatic bracket ({}, (), [], …) completion still works correctly?

Issue #1294, opened 2021-12-12
Upgrade Firebase to 9.*.*


mfix22:
Follow this upgrade guide to upgrade Firebase to version 9: https://firebase.google.com/docs/web/modular-upgrade

github4maninder:
I would like to work on this request. kindly assign this to me. thanks

github4maninder:
Hi, I would love to contribute and work on this issue. Kindly assign this to me. Thanks

github4maninder:
Hi, update is now complete. should i do a pull request. I did see there is already a pull request for eslint. should i wait for that pull request to finish.

CheRayLiu:
Hi @mfix22 could I get assigned to this? Thanks!

mfix22:
> Hi @mfix22 could I get assigned to this? Thanks!

Sure, but please don't feel the need to get assigned in order to submit PRs to fix issues. Anyone is allowed to submit a PR, even without assignment.

PR #1406, opened 2022-08-29
firebase upgrade to version 9.*.*

WOLFIEEEE:
Closes #

### DIFF
--- a/components/AuthContext.js
+++ b/components/AuthContext.js
@@ -1,5 +1,6 @@
 import React from 'react'
-import firebase from '../lib/client'
+import { getApps} from "firebase/app";
+import { getAuth,onAuthStateChanged} from 'firebase/auth';
 // IDEA: just read from firebase store at request time?
 import { client } from '../lib/api'
 
@@ -13,9 +14,12 @@ function AuthContext(props) {
   const [user, setState] = React.useState(null)
 
   React.useEffect(() => {
-    if (firebase) {
-      firebase.auth().onAuthStateChanged(newUser => setState(newUser))
+
+    if(getApps().length > 0)
+    {
+    onAuthStateChanged(getAuth() , newUser => setState(newUser))
     }
+    
   }, [])
 
   React.useEffect(() => {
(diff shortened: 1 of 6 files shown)


WOLFIEEEE:
Solving issue #1294

WOLFIEEEE:
@mfix22 hey can you please review it?

lib/client.js
```diff
@@ -1,38 +1,48 @@
...
-  const provider = new firebase.auth.GithubAuthProvider()
+  const provider = new GithubAuthProvider()
   provider.setCustomParameters({
     allow_signup: 'true',
   })
   return login(provider)
 }
 
-export default firebase.apps.length ? firebase : null
+export default FirebaseApp
```
mfix22, line 48: Can we still make this:
```js
getApps().length ? firebase : null
```
so that we don't need to call `getApps()` in other files?

components/LoginButton.js
```diff
@@ -48,7 +49,7 @@ function Drawer(props) {
 function LoginButton({ isVisible, toggleVisibility }) {
   const user = useAuth()
 
-  if (!firebase) {
+  if (getApps().length === 0) {
```
mfix22, line 52: If we make the other change suggested below, we don't need to make this change

components/AuthContext.js
```diff
@@ -13,9 +14,12 @@ function AuthContext(props) {
   const [user, setState] = React.useState(null)
 
   React.useEffect(() => {
-    if (firebase) {
-      firebase.auth().onAuthStateChanged(newUser => setState(newUser))
+
+    if(getApps().length > 0)
```
mfix22, line 18: Same comment as below.

mfix22 (changes requested):
Thanks so much for working on this @WOLFIEEEE, and so sorry for the incredible delay—I've been very busy. 

Can you please run prettier against all the changes? It's difficult to assess the diff without that 🙂 

Also leaving a few other nits for review, but besides that this LGTM.

Again, thanks so much!

Issue #1399, opened 2022-07-23
Add an option to display the programming language in the image


lokicalmito:
**Is your feature request related to a problem? Please describe.**
It's not really a problem, but it would be cool to just see what programming language that code was written into.

**Describe the solution you'd like**
A toggle in the settings panel or in the top bar which says "Display (programming) language" or something like that. 

**Describe alternatives you've considered**
An alternative would be for the user to type the language in a comment

**Additional context**
![image](https://user-images.githubusercontent.com/107685698/180618915-4eea3f72-971d-46ed-9754-60c1bd5a2877.png)

noobyogi0010:
Hey, @lokicalmito 
I would like to work on this issue if no one else has taken it up.
Thanks!

lokicalmito:
> Hey, @lokicalmito I would like to work on this issue if no one else has taken it up. Thanks!

Sure! Go ahead.

thevinitgupta:
Hey @lokicalmito can I work on this issue?

PR #1443, opened 2023-01-13
Allowing a user to toggle showing the coding language in a screenshot

allister-grange:
This is for the issue #1399. As you can see in the images below, you can now toggle "Display Language" so that you can see what language the code is written in for the screenshot. 

I was unsure as to whether this should be on by default, to be on the safe side I have defaulted it to "false".

Please let me know if anything is amiss. 
 
image

image


mfix22:
@allister-grange this it a cool idea, but I would prefer folks just use the built in title functionality for this

image

PR #1445, opened 2023-01-25
Fixed copying of updated URL to clipboard when new snippets are saved

### LINKED ISSUE #1379: Copy Plain URL doesn't copy anything
**Describe the bug**
Copying "Plain URL" doesn't work.

**To Reproduce**

1. Open https://carbon.now.sh/
2. Select the "Plain URL" item from the copy menu:

image

3. Open a new tab and paste the URL. It pastes just the site's root URL: https://carbon.now.sh/

**Expected behavior**
A link to the snippet would get copied.

- OS [e.g. macOS, Linux, Windows, iOS]: Linux
- Browser [e.g. Chrome, Safari, Firefox]: Chrome 100

agilityhawk:
The issue was observable under Copy to Clipboard > Medium.com / IFrame / Plain URL. 

- All three of these Components use the [CopyEmbed](https://github.com/carbon-app/carbon/blob/c108a251fd9edaba88638b46157421340e0b342f/components/CopyMenu.js#L107) component underneath that makes us of the [asPath](https://github.com/carbon-app/carbon/blob/c108a251fd9edaba88638b46157421340e0b342f/components/CopyMenu.js#L38) property of the nextrouter. 
- This property was not getting updated because we were using [window.history.pushState](https://github.com/carbon-app/carbon/blob/c108a251fd9edaba88638b46157421340e0b342f/components/EditorContainer.js#L63) instead of the router methods (push/replace).
- Have used [router.push](https://nextjs.org/docs/api-reference/next/router#routerpush) here since this gets invoked only when  a new snippet is created.

Caveat: Since there is a single page reload when the new snippet is created now, the page scrolls back to the top. I believe this should be a minor trade-off for making the Copy functionality work as expected.

Closes #1379


agilityhawk:
(bump). Any chance to getting this reviewed / taking it forward?

agilityhawk:
@mfix22 review/merge please?

mfix22 (approved):

PR #1488, opened 2023-08-17
Fix WidthHandler not following mouse position

awsr:
This fixes the editor edge not lining up with the mouse when adjusting the width.

Due to the editor being centered, the adjustments have to be done in increments of 2px. While 1px adjustments can be done in the settings dropdown, I have also included a fallback for the old behavior by holding `Shift` while dragging. Since this is non-obvious, I have included a brief tooltip when hovering over the width handle.


mfix22:
I think the current behavior is fine as is and we don't need to add this complexity. Thanks!

PR #1531, opened 2024-08-02
Add Docker support and update README.md

### LINKED ISSUE #1529: Dockerfile for running locally?
I would like to use carbon.now.sh at work, but I don't want to send sensitive data to your servers. I noticed you can run locally with yarn and npm, I'm wondering if you could set up a Dockerfile that runs it locally without needing to install these tools?

xufeiranfree:
- [ ] Integration tests (if applicable)

Closes #1529

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Use Node.js official image as the base image
+FROM node:18
+
+# Set the working directory inside the container
+WORKDIR /app
+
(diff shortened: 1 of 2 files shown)


knzcx:
It runs locally for me, thank you very much! Is there a way to deploy the app so that it is not running via yarn dev?

mfix22:
We aren't interested in including a Dockerfile at this time.

PR #1591, opened 2025-10-29
Add Terraform support

diofeher:
Terraform is already supported by `highlight.js`, so it's just a matter of adding it to the `Languages` list.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,4 +1,4 @@
- import toHash from 'tohash'
+import toHash from 'tohash'
 
 export const FONTS = [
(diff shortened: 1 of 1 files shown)


diofeher:
@mfix22 can you please merge this one? :D

ApprenticeofEnder:
Was legit just looking into making a PR for this! Thanks for beating me to the punch @diofeher

diofeher:
@ApprenticeofEnder The bad thing is that repo looks abandoned :(

ApprenticeofEnder:
Yeah, and CodeMirror5 et al. would need some serious finagling to get support for languages like Terraform. The versions are all fairly old.

